PR #5, opened 2015-12-07
GetAnyAgent returns random agent

dankraw:
...and some minor changes.

### DIFF
--- a/consul/agents.go
+++ b/consul/agents.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	log "github.com/Sirupsen/logrus"
 	consulapi "github.com/hashicorp/consul/api"
+	"math/rand"
 	"net/http"
 	"sync"
 )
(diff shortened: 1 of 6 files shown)


consul/agents.go
```diff
@@ -31,38 +33,48 @@ func (a *ConcurrentAgents) GetAnyAgent() (*consulapi.Client, error) {
...
 	return nil, fmt.Errorf("No agent available")
 }
 
-func (a *ConcurrentAgents) GetAgent(agentAddress string) (*consulapi.Client, error) {
+func (a *ConcurrentAgents) getRandomAgentHost() string {
+	hosts := []string{}
+	for host, _ := range a.agents {
+		hosts = append(hosts, host)
+	}
+	rand.Seed(time.Now().UTC().UnixNano())
```
janisz: Seed is set every time this function is called. Better set it once in other case we could always end up with same host

dankraw: I've removed the seeding line as it should work good enough by default.

janisz:
You can add more tests for to check error handling

consul/consul_test.go
```diff
@@ -55,23 +55,81 @@ func TestRegisterServices(t *testing.T) {
...
+
+	// given
+	server.AddService("serviceA", "passing", []string{"marathon"})
+
+	// when
+	server.Stop()
+	err := consul.Deregister("serviceA", server.Config.NodeName)
+
+	// then
+	assert.NotNil(t, err)
```
janisz: `assert.Error` will be better here

PR #11, opened 2015-12-11
separating web package

dankraw:


main.go
```diff
@@ -36,9 +37,9 @@ func main() {
 	go sync.StartSyncServicesJob(config.Sync.Interval)
 
 	// set up routes
-	http.HandleFunc("/health", HealthHandler)
-	forwarderHandler := &ForwardHandler{service, remote}
-	http.HandleFunc("/events", forwarderHandler.Handle)
+	http.HandleFunc("/health", web.HealthHandler)
+	forwardHandler := web.NewForwardHandler(service, remote)
```
janisz: `EventHandler` will be better name for it

dankraw: good point. done.

PR #12, opened 2015-12-11
Configure log format

janisz:


config/config.go
```diff
@@ -32,7 +36,13 @@ func New() (*Config, error) {
 	}
 	flag.Parse()
 	err := config.loadConfigFromFile()
-	config.setLogLevel()
+
+	if err != nil {
+		return config, err
```
dankraw: I think we should return `nil, err` here as config is already defined as `Config{}`

janisz: Fixed

dankraw:
:+1:

janisz:
We have Marathon password leak in logs. I fixed it and also make logs more consistent every log (if possible) has `"Id"` field which contains Task ID or (if not exist in current context) Application ID. Also added some more debug logs.

config/config.go
```diff
@@ -68,8 +82,11 @@ func (config *Config) parseFlags() {
 	flag.DurationVar(&config.Metrics.Interval, "metrics-interval", 30*time.Second, "Metrics reporting interval")
 	flag.StringVar(&config.Metrics.Addr, "metrics-location", "", "Graphite URL (used when metrics-target is set to graphite)")
 
+	// Log
+	flag.StringVar(&config.Log.Level, "log-level", "info", "Log level: panic, fatal, error, warn, info, or debug")
+	flag.StringVar(&config.Log.Format, "log-format", "text", "Log format: JSON, text")
```
dankraw: in code we check against "TEXT" (uppercase) - we should stick to the rule here.

PR #14, opened 2015-12-15
Hostname to Agent-address refactoring

dankraw:
We should not use Consul agent "node name" as a resolvable hostname, but rather stick to its Address.


janisz:
:+1:

PR #17, opened 2015-12-16
Create alias types for Id

janisz:
Currently our code is [Stringly Typed](http://c2.com/cgi/wiki?StringlyTyped). I added ID alias type for strings.

### DIFF
--- a/tasks/id.go
+++ b/tasks/id.go
@@ -1,0 +1,24 @@
+package tasks
+
+import "strings"
+
+// Marathon Task Id
+// Usually in the form of AppId.uuid with '/' replaced with '_'
+type Id string
+
+func (id Id) String() string {
+	return string(id)
+}
+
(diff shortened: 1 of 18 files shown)


tasks/id.go
```diff
@@ -0,0 +1,18 @@
...
+// Marathon Application Id (aka PathId)
+// Usually in the form of /rootGroup/subGroup/subSubGroup/appId
+// allowed characters: lowercase letters, digits, hyphens, slash
+type AppId string
+
+func (id AppId) String() string {
+	return string(id)
+}
+
+func (id Id) String() string {
```
dankraw: moving it closer to the type definition (just below it) would make it nicer to read. :heart:

dankraw:
great!
:+1:

tasks/id.go
janisz: Fixed

PR #23, opened 2016-01-08
Event-based deregistration fixed (now Consul-aware)

### LINKED ISSUE #21: App termination handler should check Consul service's "marathon" tag
When the app is terminated, we don't know whether or not the App has `consul:true` label because it already does not exist, so instead there has to be a request to Consul made with a 'marathon' tag check.

dankraw:
Fixes #21 

There are some gentle changes in the behaviour of event handlers, 
for example: we don't return 400 if there are no services registered in Consul etc (they could be removed by other events or it's just some marathon-noise). 

(sorry for the blank lines added :P :cookie: :cake: :cherries: )

### DIFF
--- a/consul/consul.go
+++ b/consul/consul.go
@@ -9,6 +9,7 @@ import (
 
 type ConsulServices interface {
 	GetAllServices() ([]*consulapi.CatalogService, error)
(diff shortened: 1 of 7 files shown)


janisz:
:+1:

PR #24, opened 2016-01-11
Update sample config file

janisz:


debian/config.json
```diff
@@ -16,7 +16,10 @@
     "Listen": ":4000"
   },
   "Sync": {
-    "Interval": 900000000000
+    "Enabled": false,
```
dankraw: it's `true` by default

dankraw:
:+1:

PR #25, opened 2016-01-11
Sync optimization

### LINKED ISSUE #22: Get rid of unnecessary (de)registration requests during scheduled sync
We should rewrite the sync job so we could get rid of some unnecessary request to Consul. This should  improve our metrics btw.
What is more, the stored Consul-agents list could be generated from existing Consul services (not Marathon apps) and cleaned up when needed.

dankraw:
fixes #22

### DIFF
--- a/consul/agents.go
+++ b/consul/agents.go
@@ -68,7 +68,7 @@ func (a *ConcurrentAgents) addAgent(agentHost string, agent *consulapi.Client) {
 
 func (a *ConcurrentAgents) createAgent(host string) (*consulapi.Client, error) {
 	if host == "" {
-		return nil, fmt.Errorf("Invalid addres for Agent")
+		return nil, fmt.Errorf("Invalid address for Agent")
 	}
 	config := consulapi.DefaultConfig()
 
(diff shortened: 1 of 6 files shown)


janisz:
Can you add test case for:

> What is more, the stored Consul-agents list could be generated from existing Consul services (not Marathon apps) and cleaned up when needed.

dankraw:
Well, actually it's not implemented. The assumption was wrong btw, as we cannot get agents from Consul as we don't know how access the first Consul agent. :) During sync I just prefetch agent nodes locations from application instances that are appropriately labeled. As we talked already the agents list is not cleared, as that's not an issue right now.

The main point of the issue is implemented: there are less requests to Consul during sync and logs should be more accurate.

janisz:
:+1:

dankraw:
Basically, this is the improved part:

```go
for _, task := range app.Tasks {
	if _, ok := consulServicesIdsSet[task.ID]; !ok {
		// [...]
		err := s.service.Register(service)
		// [...]
	} else {
		log.WithField("Id", task.ID).Debug("Task already registered in Consul")
	}
}
```

plus getting the agents - so sync at anytime should be successful (without waiting for Marathon events):

```go
s.addAgentNodes(apps)
```

janisz:
Now everything is clear. Read to merge

PR #30, opened 2016-01-15
Configurable Consul tag name

### LINKED ISSUE #29: Adding configurable tag name to Consul servies when registering
We should allow setting a configurable tag name instead of "marathon" for all Consul services. 
This will allow coexist multiple mesos/marathon clusters ("marathon-cluster1", "marathon-cluster2") with a single consul cluster. 
In other words, single marathon-consul instance will filter and manage only its Consul services.

dankraw:
`consul-tag` property added - fixes #29


config/config.go
```diff
@@ -62,6 +62,7 @@ func (config *Config) parseFlags() {
 	flag.StringVar(&config.Consul.SslCert, "consul-ssl-cert", "", "Path to an SSL client certificate to use to authenticate to the Consul server")
 	flag.StringVar(&config.Consul.SslCaCert, "consul-ssl-ca-cert", "", "Path to a CA certificate file, containing one or more CA certificates to use to validate the certificate sent by the Consul server to us")
 	flag.StringVar(&config.Consul.Token, "consul-token", "", "The Consul ACL token")
+	flag.StringVar(&config.Consul.Tag, "consul-tag", "marathon", "Common tag name added to every service registered in Consul, should be unique for every Marathon-cluster connected to Consul")
```
janisz: Maybe add this field in sample config?

consul/consul.go
```diff
@@ -101,7 +107,8 @@ func (c *Consul) Register(service *consulapi.AgentServiceRegistration) error {
 	return err
 }
 
-func (c *Consul) register(service *consulapi.AgentServiceRegistration) error {
+func (c *Consul) register(task *tasks.Task, app *apps.App) error {
+	service := c.marathonTaskToConsulService(task, app.HealthChecks, app.Labels)
```
janisz: I think this should be in `Register` and this function tak care only for comunicating with consul

config/config.go
dankraw: `debian/config.json`? it's added already

consul/consul.go
dankraw: good point

dankraw: fixed

janisz:
:+1:

PR #31, opened 2016-01-19
Custom service name

### LINKED ISSUE #27: Specify custom service name
I would like to be able to specify custom service name, instead of using marathon app id. Currently you need to add `consul:true` label to be visible to `marathon-consul`. We could change this syntax to: `consul:custom_service_name`. `true` could be a *protected keyword* for when you don't want any custom name (and to maintain backwards compatibility).

dankraw:
Fixes #27 

An important change has been introduced: `app_terminated_event` is no more being handled, `deployment_info` and `deployment_step_success` handling added instead.


consul/consul.go
```diff
@@ -13,9 +12,9 @@ import (
 
 type ConsulServices interface {
 	GetAllServices() ([]*consulapi.CatalogService, error)
-	GetServices(name tasks.AppId) ([]*consulapi.CatalogService, error)
-	Register(task *tasks.Task, app *apps.App) error
-	Deregister(serviceId tasks.Id, agentAddress string) error
+	GetServices(name string) ([]*consulapi.CatalogService, error)
+	Register(task *apps.Task, app *apps.App) error
```
janisz: I think we can get rid of `task` parameter since it could be embedded in `app`

dankraw: this will end up with a big code refactor, I would suggest doing this in the future. :)

apps/app.go
```diff
@@ -26,13 +26,36 @@ type AppsResponse struct {
...
 
 func (app *App) IsConsulApp() bool {
-	value, ok := app.Labels["consul"]
-	return ok && value == "true"
+	_, ok := app.Labels["consul"]
+	return ok
+}
+
+func (app *App) ConsulServiceName() string {
+	if value, ok := app.Labels["consul"]; ok && value != "true" {
```
janisz: We should put warning that application will not be rgistered as 'true'

web/event_handler.go
```diff
@@ -234,6 +191,61 @@ func (fh *EventHandler) handleStatusEvent(w http.ResponseWriter, body []byte) {
 	}
 }
 
+func (fh *EventHandler) handleDeploymentInfo(w http.ResponseWriter, body []byte) {
```
janisz: Custom service name should be also handled here. Consider following situation. 
1. Deployed app `A` with label `consul=true`
2. Changed app label to `consul=B`

You end up with two services in consul: A and B.

web/event_handler.go
```diff
@@ -234,6 +191,61 @@ func (fh *EventHandler) handleStatusEvent(w http.ResponseWriter, body []byte) {
...
+		fh.handleBadRequest(err, w)
+		return
+	}
+
+	log.Info("Got DeploymentInfoEvent")
+
+	errors := []error{}
+
+	for _, app := range deploymentInfo.StoppedApps() {
+		if app.IsConsulApp() {
```
janisz: Body of this statement should be extracted to function

apps/app.go
dankraw: The behaviour is described in the README. How would you see the warning implemented - as an additional log?

janisz: Yep, I was thinking about log and explicite mark this as deprecated and remove in next version

consul/consul.go
janisz: OK

apps/app.go
dankraw: Added a warning to Consul.Register()

web/event_handler.go
dankraw: fixed by handling `deployment_step_success` for Restarted and renamed apps

dankraw: thanks!

web/event_handler.go
dankraw: done

events/deployment_info.go
```diff
@@ -0,0 +1,122 @@
...
+
+func (d *DeploymentEvent) appsMap(applications []*apps.App) map[apps.AppId]*apps.App {
+	result := make(map[apps.AppId]*apps.App)
+	for _, app := range applications {
+		result[app.ID] = app
+	}
+	return result
+}
+
+func (d *DeploymentEvent) consulAppsWithActionPerformed(allApps []*apps.App, actionType string) []*apps.App {
```
janisz: Nice

events/deployment_info.go
```diff
@@ -0,0 +1,122 @@
...
+	if d.CurrentStep != nil {
+		if d.CurrentStep.Actions != nil {
+			return d.CurrentStep.Actions
+		}
+	}
+	return []*Action{}
+}
+
+func (d *DeploymentEvent) StoppedConsulApps() []*apps.App {
+	return d.consulAppsWithActionPerformed(d.OriginalApps(), "StopApplication")
```
janisz: Maybe we shoud extract actions to some const and link them with Marathon code https://github.com/mesosphere/marathon/blob/e3e867f00d4964d0367ec7e07519d6c9545681d6/src/main/scala/mesosphere/marathon/upgrade/DeploymentPlan.scala#L19-L34

janisz:
LGTM but next time make smaller PR.
@adamdubiel what do you think?

events/deployment_info.go
dankraw: I've linked our readme to the Marathon EventBus docs.

I was trying to implement these string literals as constants but it didn't make the code more readable at all (`events.STOP_APPLICATION_ACTION` for example). They appear only in this file (the API is with private scope) plus one more occurrence for both actions in a single test (for event creation purpose), so maybe let's just not do this. :) Or maybe you have an idea for a better implementation...

dankraw:
@janisz This PR was meant to be more concise, sorry. ;)

adamdubiel:
+1 on smaller PRs as a rule, but i don't feel entitled to give opinion about size of this particular one

janisz:
@adamdubiel What do you think about code?

dankraw:
@janisz 
IMHO @adamdubiel is focused on the feature being delivered, not reviewing these 1.5k lines of code. ;) I'm gonna merge it as it's already tested. We will hire someone else from Skylab for the upcoming issues to be resolved, so the reviewers team will grow. :)

PR #40, opened 2016-01-27
Fixing deployment events handling for apps in groups

### LINKED ISSUE #35: Not handling properly deployment events for apps with custom parent groups in Marathon
Currently `RestartedConsulApps`/`StoppedConsulApps`/`RenamedConsulApps` api of deployment events search for the apps in `/Plan/Original/Apps` of the event - this is will catch only apps that are created in root of the deployment plan. 
When application has a different parent group, for example `/com.example.tech/applicationName` we should look in `/Plan/Original/Groups[{group_index}]/Apps` where `{group_index}` points a record having `id: /com.example.tech`.

dankraw:
fixes #35

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,16 @@ Endpoint  | Description
 `/health` | healthcheck - returns `OK`
 `/events` | event sink - returns `OK` if all keys are set in an event, error message otherwise
 
+### Known limitations
+
+The following section describes known limitations with `marathon-consul`.
+
+* Every marathon application needs to have a unique service name in Consul.
+* In Marathon when a deployment changing the application's service name (by changing its `labels`) is being stopped, it changes app's configuration anyway.
+  This means we loose the link between the app and the services registered with the old name in Consul. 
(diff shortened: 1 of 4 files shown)


janisz:
Custom names looks for the first sight looks easy, but now we get more and more complicated code. There could be more corner cases than we previously thought. We need to discuss this feature.
I'll test this PR tomorrow, code looks good :+1:

adamdubiel:
Maybe we should pivot now and make it as simple as possible. @dankraw suggested some simple k/v mapping, Consul has own k/v so why not use it if we go this way.

dankraw:
OK, so as we spoke, we're gonna keep the current implementation for v0.3.x and maybe return to the KV-store idea in the future if there are more issues with parsing large events / weird corner-cases etc.
The KV-store integration is not cheap either as we thought at first. The data has to be put there depending on some stimulus/app configuration, it needs to be replicated to all datacenters, plus we would have to rethink the how to support renaming of services.

PR #43, opened 2016-01-28
Service host to IPv4 resolving

### LINKED ISSUE #41: Using IP address instead of hostname as ServiceAddress
marathon-consul uses FQDN as a ServiceAddress instead of IP address when registering in Consul. Using FQDN makes no sense here.

dankraw:
fixes #41


janisz:
:+1:

Issue #49, opened 2016-02-04
Event-sourced requests to Consul should not be performed using random Agent


dankraw:
Example scenario:
* we get TASK_FAILED
* we check if there is a service registered with the same ID and properly labeled -> we call GetServices using random agent that may be down -> we have an error, the service is not being deregistered.

We should use random agents only for sync, we could remove invalid agents from cache.

PR #50, opened 2016-02-04
Getting Consul services with retries + agents cache cleanup

dankraw:
Some changes introduced:
* getting services from Consul with retries on failure (using another random agent)
* failing agents get removed from cache
* agents cache size metric added

solves #49


consul/agents.go
```diff
@@ -21,6 +24,8 @@ type ConcurrentAgents struct {
 	lock   sync.Mutex
 }
 
+var consulAgentsCacheSizeMetricName string = "consul.agents.cache.size"
```
janisz: `const`

dankraw: fixed

consul/agents.go
```diff
@@ -28,14 +33,15 @@ func NewAgents(config *ConsulConfig) *ConcurrentAgents {
 	}
 }
 
-func (a *ConcurrentAgents) GetAnyAgent() (*consulapi.Client, error) {
+func (a *ConcurrentAgents) GetAnyAgent() (*consulapi.Client, string, error) {
```
janisz: IMO it should be documented or named what second argument is.

consul/agents.go
```diff
@@ -99,13 +127,15 @@ func (a *ConcurrentAgents) createAgent(host string) (*consulapi.Client, error) {
...
-		"TokenEnabled": a.config.Token != "",
+		"Address":                config.Address,
+		"Scheme":                 config.Scheme,
+		"Timeout":                config.HttpClient.Timeout,
+		"BasicAuthEnabled":       a.config.Auth.Enabled,
+		"TokenEnabled":           a.config.Token != "",
 		"SslVerificationEnabled": a.config.SslVerify,
 	}).Debug("Creating Consul client")
 
+	metrics.IncrementGauge(consulAgentsCacheSizeMetricName)
```
janisz: IMO this should be near `addAgent`

consul/consul.go
```diff
@@ -30,10 +33,29 @@ func New(config ConsulConfig) *Consul {
...
-	if err != nil {
-		return nil, err
+	return c.getServicesUsingProviderWithRetriesOnAgentFailure(func(agent *consulapi.Client) ([]*consulapi.CatalogService, error) {
+		return c.getServicesUsingAgent(name, agent)
+	})
+}
+
+func (c *Consul) getServicesUsingProviderWithRetriesOnAgentFailure(provide ServicesProvider) ([]*consulapi.CatalogService, error) {
+	var services []*consulapi.CatalogService = nil
+	for {
```
janisz: Shouldn't we limit retires?

consul/consul.go
```diff
@@ -163,7 +185,7 @@ func (c *Consul) GetAgent(agentAddress string) (*consulapi.Client, error) {
 }
 
 func (c *Consul) marathonTaskToConsulService(task *apps.Task, app *apps.App) (*consulapi.AgentServiceRegistration, error) {
-	IP, err := task.HostToIPv4()
+	IP, err := utils.HostToIPv4(task.Host)
```
janisz: :+1:  good move

consul/agents.go
dankraw: right, added names to the interface (implementation left as is).

consul/consul.go
dankraw: why? :) 
They're limited to the size of agents cache. Usually a single failing agent should be enough.

consul/agents.go
```diff
@@ -47,18 +51,45 @@ func (a *ConcurrentAgents) getRandomAgentHost() string {
...
+func (a *ConcurrentAgents) RemoveAgent(agentAddress string) {
 	a.lock.Lock()
 	defer a.lock.Unlock()
-	if agent, ok := a.agents[agentHost]; ok {
+
+	if IP, err := utils.HostToIPv4(agentAddress); err != nil {
+		log.WithError(err).Error("Could not remove agent from cache")
+	} else {
+		agentAddress = IP.String()
+		if _, ok := a.agents[agentAddress]; ok {
```
janisz: IMO this condition is redundant.

> [The delete function doesn't return anything, and will do nothing if the specified key doesn't exist.](https://blog.golang.org/go-maps-in-action)

consul/agents.go
```diff
@@ -47,18 +51,45 @@ func (a *ConcurrentAgents) getRandomAgentHost() string {
...
+
+func (a *ConcurrentAgents) GetAgent(agentAddress string) (*consulapi.Client, error) {
+	a.lock.Lock()
+	defer a.lock.Unlock()
+
+	IP, err := utils.HostToIPv4(agentAddress)
+	if err != nil {
+		return nil, err
+	}
+	agentAddress = IP.String()
```
janisz: Maybe `HostToIPv4` should return string or (better idea) we should use `IP` everywhere we can to have clear situation what is stored in `agentAddress`. Another thing overriding parameters value is considered as a bad idea, it is difficult to read because `agentAddress` we are using here is not `agentAddress` that was passed

consul/consul.go
janisz: Sounds good

consul/agents.go
dankraw: Yep, I know, but we had the "decrement gauge" metric being updated before, so it was intended. Now, this can be simplified, as we always pass the actual cache size on update.

consul/agents.go
dankraw: I like the idea, but I'm quite worry this will mean lots of api changes. :> Will try look into this.

dankraw: I would like to perform the refactor in another PR, how about that?

consul/agents.go
dankraw: fixed

consul/agents.go
dankraw: Unfortunately, we can't have net.IP as a map key. We could wrap string, but I think we can leave it for future refactoring.

janisz:
Looks good to me

PR #60, opened 2016-02-19
Fixes #55 | Smart agents cache cleanup

### LINKED ISSUE #55: Agents cache smart cleanup
We should not remove agents from cache after first request failure. We might get `Unexpected response code: 500 (rpc error: rpc error: No cluster leader)` and few seconds later the clients works fine. 

An agent should be removed from cache after a specified number of consecutive failures (timeout, connection refused), if it responds (even with status code 500) it should not be removed. We may consider doing some priority queue for search failing clients last.

janisz:


consul/consul.go
```diff
@@ -40,20 +40,25 @@ func (c *Consul) GetServices(name string) ([]*consulapi.CatalogService, error) {
...
 		}
-		if services, err = provide(agent); err != nil {
-			log.WithError(err).WithField("Address", agentAddress).Error("An error occurred getting services from Consul, retrying with another agent")
-			c.agents.RemoveAgent(agentAddress)
+		if services, err := provide(agent.Client); err != nil {
+			log.WithError(err).WithField("Address", agent.IP).
+				Error("An error occurred getting services from Consul, retrying with another agent")
+			if failures := agent.IncFailures(); failures > c.config.MaxAgentGetServicesFailures {
+				log.WithField("Address", agent.IP).WithField("Failures", failures).
+					Warn("Removing agent due too many failures")
```
dankraw: "due to too many"

consul/consul.go
```diff
@@ -40,20 +40,25 @@ func (c *Consul) GetServices(name string) ([]*consulapi.CatalogService, error) {
...
 		if err != nil {
 			return nil, err
 		}
-		if services, err = provide(agent); err != nil {
-			log.WithError(err).WithField("Address", agentAddress).Error("An error occurred getting services from Consul, retrying with another agent")
-			c.agents.RemoveAgent(agentAddress)
+		if services, err := provide(agent.Client); err != nil {
+			log.WithError(err).WithField("Address", agent.IP).
+				Error("An error occurred getting services from Consul, retrying with another agent")
+			if failures := agent.IncFailures(); failures > c.config.MaxAgentGetServicesFailures {
```
dankraw: I think this should  be `>=`, now if `maxFailures = 3`, agent is being removed after 4th failure. A test case would be helpful, if it's possible to check it.

consul/agents_test.go
```diff
@@ -88,9 +113,23 @@ func TestRemoveAgent(t *testing.T) {
...
-		assert.Equal(t, agent, agent2)
-		assert.Equal(t, "127.0.0.2", address)
+		agent, err := agents.GetAnyAgent()
+		assert.Equal(t, agent.Client, agent2)
+		assert.Equal(t, "127.0.0.2", agent.IP)
 		assert.NoError(t, err)
 	}
 }
+
+func TestDoubleRemoveAgentShouldPass(t *testing.T) {
```
dankraw: `TestRemoveAgentTwiceShouldPass` :)

config/config.go
```diff
@@ -70,6 +70,8 @@ func (config *Config) parseFlags() {
 	flag.StringVar(&config.Consul.Token, "consul-token", "", "The Consul ACL token")
 	flag.StringVar(&config.Consul.Tag, "consul-tag", "marathon", "Common tag name added to every service registered in Consul, should be unique for every Marathon-cluster connected to Consul")
 	flag.DurationVar(&config.Consul.Timeout, "consul-timeout", 3*time.Second, "Time limit for requests made by the Consul HTTP client. A Timeout of zero means no timeout")
+	flag.Uint32Var(&config.Consul.MaxAgentGetServicesFailures, "consul-max-agent-failures", 3, "Max count of failed responses for agent")
```
dankraw: How about:
`MaxAgentGetServicesFailures` -> `AgentFailuresTolerance`
and
"Max count of failed responses for agent" ->  "Max number of consecutive request failures for agent before removal from cache"

config/config.go
```diff
@@ -70,6 +70,8 @@ func (config *Config) parseFlags() {
...
+	flag.Uint32Var(&config.Consul.GetServicesRetries, "consul-get-services-retry", 3, "Retries count for failed request")
```
dankraw: Maybe:
`GetServicesRetries` -> `RequestRetries`
"Number of retries on failure when performing requests to Consul. Each retry uses different cached agent."

consul/consul.go
dankraw: On the other hand, after renaming the config property to "failure tolerance" this makes sense, 0 means no tolerance.

janisz:
Renamed Consul config fields and add explicit test for `getServicesUsingProviderWithRetriesOnAgentFailure`

dankraw:
nice! :+1:

PR #62, opened 2016-02-19
Fixes #59 | Get only consul apps from marathon

### LINKED ISSUE #59: Use marathon labels filter API
When we obtain apps from marathon we download all apps and then filter it on our side. We could use 

    GET /v2/apps?label={labelSelectorQuery}

[endpoint](https://mesosphere.github.io/marathon/docs/rest-api.html)  and download only apps that are interesting for us. [This functions](https://github.com/allegro/marathon-consul/blob/523192ee16f05993595cc1f77ebb7b91958106ed/marathon/marathon.go#L59-L78) should be changed.

janisz:


marathon/marathon.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 type Marathoner interface {
-	Apps() ([]*apps.App, error)
+	AppsWithLabel(label string) ([]*apps.App, error)
```
dankraw: Personally, I would prefer something less technical in this api, like `ManagedApps()` or `ConsulApps()` which would be implemented as private `return appsWithLabel(apps.MARATHON_CONSUL_LABEL)`. We wouldn't  have to pass that "consul" label value everywhere in the project. 
We could go further and rename `app.IsConsulApp()` to `app.IsManagedApp()` if necessary.

dankraw:
:+1: but see comment :>

janisz:
Renamed `AppsWithLabels` → `ConsulApps`

dankraw:
:+1:

PR #77, opened 2016-04-19
Unregister tasks (even for apps without label)

janisz:
Simpify task deregistration. Do not query marathon to check
if this app is managed by marathon-consul.


dankraw:
👍

PR #84, opened 2016-06-17
Fixes #83 | Unregister unhealthy killed tasks

### LINKED ISSUE #83: Support `unhealthy_task_kill_event`
Marathon 1.1.1 introduced new event [`unhealthy_task_kill_event`](https://mesosphere.github.io/marathon/docs/event-bus.html) we should support it.

Changes are required in `events` where new event struct must be added and in [`event_handler.go`](https://github.com/allegro/marathon-consul/blob/fb761a02d6bc4357ff8f618c066350730b71c3ab/web/event_handler.go#L73) where it should be handled.

janisz:
Marathon 1.1.1 introduced new event `unhealthy_task_kill_event`
that occurs when task is scheduled to kill after becomes unhealthy.
This change make marathon-consul aware of this event.

### DIFF
--- a/web/event_handler.go
+++ b/web/event_handler.go
@@ -75,6 +75,8 @@ func (fh *eventHandler) handleEvent(eventType string, body []byte) error {
 		return fh.handleStatusEvent(body)
 	case "health_status_changed_event":
 		return fh.handleHealthStatusEvent(body)
+	case "unhealthy_task_kill_event":
+		return fh.handleUnhealthyTaskKillEvent(body)
 	case "deployment_info":
 		return fh.handleDeploymentInfo(body)
 	case "deployment_step_success":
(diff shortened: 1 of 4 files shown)


web/event_handler.go
```diff
@@ -236,16 +247,25 @@ func (fh *eventHandler) deregisterAllAppServices(app *apps.App) []error {
...
 	return errors
 }
 
+func (fh *eventHandler) deregister(serviceId apps.TaskId, agentAddress string) error {
+	err := fh.service.Deregister(serviceId, agentAddress)
+	if err != nil {
+		log.WithField("Id", serviceId).WithError(err).Error("There was a problem deregistering task")
+		return err
+	} else {
+		return nil
```
dankraw: we could simply return `err` :)

dankraw:
👍

web/event_handler.go
janisz: Fixed

PR #85, opened 2016-06-17
Fixes #81 configure consul name sparator

### LINKED ISSUE #81: Consul does not support dots in service names for DNS queries
Services names must not contain a dot (.) otherwise you could not use the Consul DNS API. For example
a Marathon App with id "/demo/hello/world" is registered by default as Consul Service "demo.hello.world"

From Consul DNS API documentation (https://www.consul.io/docs/agent/services.html):
"DNS-compliant service and tag names may contain any alpha-numeric characters, as well as dashes. Dots are not supported because Consul internally uses them to delimit service tags."

It would be nice if you could implement a different naming strategy, e.g. use dashes instead of dots.

Current workaround for us is to specify a custom consul service name.

janisz:
Allow user to specify which separator will be used to create
default service name.

### DIFF
--- a/apps/app.go
+++ b/apps/app.go
@@ -8,6 +8,10 @@ import (
 // Only Marathon apps with this label will be registered in Consul
 const MARATHON_CONSUL_LABEL = "consul"
 
+// Separator used between groups and task id
+// used when creating default service name for Consul
+var CONSUL_NAME_SEPARATOR = "."
+
 type HealthCheck struct {
 	Path                   string `json:"path"`
 	PortIndex              int    `json:"portIndex"`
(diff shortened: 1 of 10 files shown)


apps/app.go
```diff
@@ -46,7 +50,7 @@ func (id AppId) String() string {
 }
 
 func (id AppId) ConsulServiceName() string {
-	return strings.Replace(strings.Trim(strings.TrimSpace(id.String()), "/"), "/", ".", -1)
+	return strings.Replace(strings.Trim(strings.TrimSpace(id.String()), "/"), "/", CONSUL_NAME_SEPARATOR, -1)
```
dankraw: Personally, I'm not sure if it's the best implementation possible (global `var` in `apps`, quite scary), maybe we should add a `separator` argument to this method and pass the configuration variable every time? Or even better, move it to Consul class and have `Consul.ServiceName(appId)`? What do you think @janisz?

janisz: Passing variable every time we need get consul name is not an option. Then in events we will need global variable. I'd prefer to move this logic to consul object but this will be a big change. Apps and Events will be affected and so this will lead to huge changes in whole codebase.

janisz:
Closing in favour of #87

PR #87, opened 2016-06-20
Configurable Consul service name separator

### LINKED ISSUE #81: Consul does not support dots in service names for DNS queries

dankraw:
Fixes #81 Different implementation, app -> service name conversion moved to Consul object.

Contains work from @janisz at #85 

👯 👯 👯


README.md
```diff
@@ -104,42 +104,41 @@ curl -X POST 'http://marathon.service.consul:8080/v2/eventSubscriptions?callback
...
+metrics-location       |                       | Graphite URL (used when metrics-target is set to graphite)
+metrics-prefix         | `default`             | Metrics prefix (resolved to `<hostname>.<app_name>` by default)
+metrics-target         | `stdout`              | Metrics destination `stdout` or `graphite` (empty string disables metrics)
+sync-enabled           | `true`                | Enable Marathon-consul scheduled sync
+sync-force             | `false`               | Force leadership-independent Marathon-consul sync (run always)
+sync-interval          | `15m0s`               | Marathon-consul sync [interval](https://golang.org/pkg/time/#Duration) **Note:** While using file configuration intervals should be provided in *nanoseconds*
+sync-leader            | `<hostname>:8080`     | Marathon cluster-wide node name (defaults to `<hostname>:8080`), the sync will run only if the node is the current [Marathon-leader](https://mesosphere.github.io/marathon/docs/rest-api.html#get-v2-leader)
+workers-pool-size      | `10`                  | Number of concurrent workers processing events
+
```
janisz: Can we minify this difference?

dankraw: Well, I didn't manually touch the file. You should know the differences better than me. :D

dankraw: OK I get it. There should be little difference. Maybe something went wrong with the rebase. :)

dankraw: done :)

apps/app.go
```diff
@@ -45,23 +44,20 @@ func (id AppId) String() string {
 	return string(id)
 }
 
-func (id AppId) ConsulServiceName() string {
-	return strings.Replace(strings.Trim(strings.TrimSpace(id.String()), "/"), "/", ".", -1)
-}
-
 func (app *App) IsConsulApp() bool {
```
janisz: If we are moving Consul related logic from here this could be also moved to consul.

apps/app.go
```diff
@@ -45,23 +44,20 @@ func (id AppId) String() string {
...
 	return ok
 }
 
-func (app *App) ConsulServiceName() string {
-	if value, ok := app.Labels[MARATHON_CONSUL_LABEL]; ok && value != "true" {
-		value = AppId(value).ConsulServiceName()
-		if value != "" {
-			return value
-		}
+func (app *App) ConsulName() string {
```
janisz: If we are moving Consul related logic from here this could be also moved to consul.

config/config.go
```diff
@@ -71,11 +71,12 @@ func (config *Config) parseFlags() {
...
 	flag.Uint32Var(&config.Consul.RequestRetries, "consul-get-services-retry", 3, "Number of retries on failure when performing requests to Consul. Each retry uses different cached agent")
+	flag.StringVar(&config.Consul.ConsulNameSeparator, "consul-name-separator", ".", "Separator used to create default service name for Consul")
 
 	// Web
-	flag.StringVar(&config.Web.Listen, "listen", ":4000", "Accept connections at this address")
-	flag.IntVar(&config.Web.QueueSize, "events-queue-size", 1000, "Size of events queue")
-	flag.IntVar(&config.Web.WorkersCount, "workers-pool-size", 10, "Number of concurrent workers processing events")
+	flag.StringVar(&config.Web.Listen, "listen", ":4000", "accept connections at this address")
+	flag.IntVar(&config.Web.QueueSize, "events-queue-size", 1000, "size of events queue")
+	flag.IntVar(&config.Web.WorkersCount, "workers-pool-size", 10, "number of concurrent workers processing events")
```
janisz: Why change it to lowercase? In [last commit](https://github.com/allegro/marathon-consul/commit/6ed7737d4e5ddf82bd078005b8caa6a6b18270b5/config/config.go) I make it uppercase to be consistent with other.

apps/app.go
dankraw: In my opinion it should stay here as it's related to the Marathon app configuration. Here `ConsulName` returns the name that the user set for the app on the Marathon side (using it's labels or simply getting the default app-id). That's why we do the `"true"` and `""` comparison here, etc. 
On the Consul class side we have the destination service name returned with all the Consul logic applied.

config/config.go
dankraw: Again, a rebase issue. :) Fixed, thanks!

apps/app.go
dankraw: Same as below, it's related to Marathon app configuration, not Consul services.

janisz:
:+1:

PR #95, opened 2016-08-10
Fix wrong tag field

janisz:
I notice that `make build` fails because two field tags are invalid. 
```
apps/app.go:19: struct field tag `json:"value` not compatible with reflect.StructTag.Get: bad syntax for struct tag value
exit status 1
consul/consul_test.go:537: struct field tag `json:"value` not compatible with reflect.StructTag.Get: bad syntax for struct tag value
exit status 1
make: *** [test] Error 1
```
I fixed it and change CI to check whole project builds not only tests pass.


dankraw:
👍

xor-xor:
LGTM

pbetkier:
👍

PR #96, opened 2016-08-10
Enable race detector in tests

janisz:
Enable `-race` during tests and fix race conditions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 PACKAGES=$(shell go list ./... | grep -v /vendor/)
-TESTARGS?=
+TESTARGS?=-race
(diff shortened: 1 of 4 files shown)


xor-xor:
LGTM

Makefile
```diff
@@ -1,5 +1,5 @@
 TEST?=./...
-TESTARGS?=
+TESTARGS?=-race
```
pbetkier: Is it `-race` or `--race`?

janisz: Golang has it's own model for flags
https://groups.google.com/d/msg/golang-nuts/3myLL-6mA94/VUkLtSOyS-YJ

Issue #89, opened 2016-07-06
Support multiple ports


avthart:
Most of our marathon apps have only one portMapping but sometimes we have an app with multiple ports where one of the ports will be used for (http) health checks. 

In newer versions of marathon you can specify labels for each port in the "portMappings". Maybe this is something we can use to register services in consul?

Example marathon app:
```
{
  "id": "/hello-world",
  "cpus": 0.1,
  "mem": 256,
  "instances": 1,
  "container": {
    "type": "DOCKER",
    "docker": {
      "image": "my-hello-world:latest",
      "network": "BRIDGE",
      "portMappings": [
        {
          "containerPort": 8080,
          "hostPort": 0,
          "protocol": "tcp",
          "name": "api",
          "labels": { 
              "consul": "helloworld-api"
          }
        },
        {
          "containerPort": 8081,
          "hostPort": 0,
          "protocol": "tcp",
          "name": "actuator",
          "labels": {
              "consul": "helloworld-health"
          }
        }
      ]
   }
  },
  "healthChecks": [
    {
      "path": "/health",
      "protocol": "HTTP",
      "portIndex": 1,
      "gracePeriodSeconds": 300,
      "intervalSeconds": 10,
      "timeoutSeconds": 10,
      "maxConsecutiveFailures": 3,
      "ignoreHttp1xx": false
    }
  ]
}
```

janisz:
This could be great addition to marathon-consul. Internally we don't use docker containers a lot, so it's no priority for us to support docker port labels. 
Although if you need this feature we can help you to deliver it. Changes are required in [`consul.marathonTaskToConsulService`](https://github.com/allegro/marathon-consul/blob/master/consul/consul.go#L210). I recommend do it step by step:

- [ ] Detect which port should be registered basde on port labels (in no one specified 0 wil be used)
- [ ] Merge port labels with app labels (How to handle conflicts?)
- [ ] Generate multiple consul services from one marathon task

If you need any guidance please reach us here or with e-mail/twitter/hangout so we can help you.

pbetkier:
It appears we have a similar need now internally – registering an app under more than 1 name with different ports, but with the same healthcheck. It's not related to docker though. I'm coding such solution right now, will open a PR for discussion in a couple of days.

karthimohan:
@pbetkier Thanks for merging #99 Just wondering if you have any timeline on when this #89 Enhancement will be available for testing.

pbetkier:
[...]

PR #99, opened 2016-08-12
Separate service id from task

pbetkier:
Consul service used to be registered under the same id as its related marathon task id. This PR separates these two ids. Marathon task id is remembered in service's `marathon-task` tag. Also did some refactorings: extracted Service and ServiceId domain types, and also ServiceRegistry interface to make things more clear by hiding Consul internals.

I'm still working on this PR, going to cleanup some code and verify that all branches are covered with tests. But the main idea is there, so waiting for your feedback.

It's a prerequisite to #89, because we would register more than one service for a given marathon task.


janisz:
I think this should be split into 2 PR: introduce Service and ServiceId domain types and change ServiceId to generated uuid. This will make review and testing easier. Another thing changing IDs could be harmful so we need to test impact of this action.

pbetkier:
Even if do we split, we have to test this PR thoroughly because it's a significant change. Why not do it once? I've split the change into two separate commits for the review's sake, but I don't see the point in splitting it into separate PRs. One would contain the feature, but introducing complexity in the codebase and the other would only clean things up. Still, we merge either both or nothing IMO :)

janisz:
I thought we can merge refactoring first and then ID change. With refactoring most of tests should remain untouched and by this we will ensure change doesn't break anything.

janisz:
I created dedicated task (#100) for extracting domain service object. It should be backward comaptible and we could release it as 4.1 build with Go 1.7.

service/service.go
```diff
@@ -0,0 +1,44 @@
+package service
+
+import (
+	"errors"
+	"strings"
+
+	"github.com/allegro/marathon-consul/apps"
+	"fmt"
```
janisz: `find ./ -name '*.go' | xargs goimports -w`

service/service.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+type Service struct {
+	ID                      ServiceId
+	Name                    string
+	Tags                    []string
+	RegisteringAgentAddress string
+}
+
+func (s *Service) TaskId() (apps.TaskId, error) {
```
janisz: Why pointer receiver?

consul/consul.go
```diff
@@ -63,58 +56,71 @@ func (c *Consul) getServicesUsingProviderWithRetriesOnAgentFailure(provide Servi
...
+				for _, consulServiceInstance := range consulServiceInstances {
+					allInstances = append(allInstances, c.consulServiceToService(consulServiceInstance))
+				}
 			}
 		}
 	}
 	return allInstances, nil
 }
 
+func (c *Consul) consulServiceToService(consulService *consulapi.CatalogService) *service.Service {
```
janisz: How about moving this function to `service` package and make it constructor function e.g. `func FromConsul(consulService *consulapi.CatalogService) *service.Service`

consul/consul.go
```diff
@@ -1,32 +1,25 @@
...
 
 import (
+	"errors"
 	"fmt"
 	"net/url"
 	"strings"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/allegro/marathon-consul/apps"
+	"github.com/allegro/marathon-consul/service"
```
janisz: Consul package shouldn't have dependency to service.

consul/consul.go
```diff
@@ -162,9 +168,32 @@ func (c *Consul) register(service *consulapi.AgentServiceRegistration) error {
 	return err
 }
 
-func (c *Consul) Deregister(serviceId apps.TaskId, agentAddress string) error {
+func (c *Consul) DeregisterByTask(taskToDeregister apps.TaskId, agentAddress string) error {
```
janisz: `taskToDeregister` → `taskIDToDeregister` or just `taskID`

consul/consul.go
pbetkier: Then service package would depend on consul package, not cool :)

consul/consul.go
pbetkier: I'd say right the opposite, you don't want our domain types with business logic depend on infrastructure, external classes like consul.

janisz: Good point

consul/consul.go
janisz: OK. But this could be a pure function not depended on `*Consul`

consul/consul.go
```diff
@@ -35,24 +28,24 @@ func New(config ConsulConfig) *Consul {
...
+func (c *Consul) getServicesUsingProviderWithRetriesOnAgentFailure(provide ServicesProvider) ([]*service.Service, error) {
 	for retry := uint32(0); retry <= c.config.RequestRetries; retry++ {
 		agent, err := c.agents.GetAnyAgent()
 		if err != nil {
 			return nil, err
 		}
 		if services, err := provide(agent.Client); err != nil {
 			log.WithError(err).WithField("Address", agent.IP).
-				Error("An error occurred getting services from Consul, retrying with another agent")
+			Error("An error occurred getting services from Consul, retrying with another agent")
```
medzin: Are you sure it is Error and not a Debug/Notice message?

consul/consul.go
```diff
@@ -162,9 +168,32 @@ func (c *Consul) register(service *consulapi.AgentServiceRegistration) error {
...
 }
 
-func (c *Consul) Deregister(serviceId apps.TaskId, agentAddress string) error {
+func (c *Consul) DeregisterByTask(taskToDeregister apps.TaskId, agentAddress string) error {
+	service, err := c.findServiceByTaskId(taskToDeregister)
+	if err != nil {
+		return err
+	}
+	c.Deregister(service)
+	return err
```
medzin: IMHO - better to be explicit here so`return nil`

consul/consul.go
pbetkier: Good point, I'll make it a pure function.

consul/consul.go
pbetkier: @janisz what do you think? Warn / Debug? Should we change levels in this PR?

janisz: I'll leave it as is (without changing indentation)

pbetkier:
@janisz @dankraw I fixed according to review comments, can I merge this into develop?

dankraw:
late but 👍  :>

PR #102, opened 2016-08-17
Fixes #101 | Use glide to lock dependecies

### LINKED ISSUE #101: Handle dependecies
Currently we do not specify version of dependencies we are using. Consider using some type of [package manager](https://github.com/golang/go/wiki/PackageManagementTools) or vendor directory to make builds independent of machine state and dependencies changes.

janisz:
Dependencies are described in glide.yaml and frozen by glide.lock. To reduce steps when building project vendor catalog is committed with all dependencies.


pbetkier:
Juicy contribution @janisz :)

pbetkier:
👍

janisz:
I'm not sure if it's a good way of storing dependencies but I can't find anything better then storing all deps in vendor dir.

PR #103, opened 2016-08-18
Tests for separating service id from task id

pbetkier:
Adds missing tests for #99


pbetkier:
coveralls is lying, we do have more coverage now, I swear!

janisz:
LGTM :+1:

dankraw:
👍

PR #104, opened 2016-08-19
More efficient deregistering by task

pbetkier:
Deregistering by task used to load all nodes (service instances) and loop over them in search for marathon task tag. With this change it loads all services and only loads nodes of services containing the marathon task tag.

It also correctly handles the corner case when there are multiple nodes containing the same marathon task id tag – it deregisters all of them. Before it would deregister only the first occurrence.

### DIFF
--- a/consul/consul.go
+++ b/consul/consul.go
@@ -57,25 +57,36 @@ func (c *Consul) getServicesUsingProviderWithRetriesOnAgentFailure(provide Servi
 }
 
 func (c *Consul) getServicesUsingAgent(name string, agent *consulapi.Client) ([]*service.Service, error) {
-	datacenters, err := agent.Catalog().Datacenters()
+	dcAwareQueries, err := dcAwareQueriesForAllDcs(agent)
 	if err != nil {
 		return nil, err
 	}
 	var allServices []*service.Service
 
-	for _, dc := range datacenters {
-		dcAwareQuery := &consulapi.QueryOptions{
-			Datacenter: dc,
-		}
+	for _, dcAwareQuery := range dcAwareQueries {
 		allConsulServices, _, err := agent.Catalog().Service(name, c.config.Tag, dcAwareQuery)
 		if err != nil {
 			return nil, err
 		}
-		for _, consulService := range allConsulServices {
-			allServices = append(allServices, consulServiceToService(consulService))
-		}
+		allServices = append(allServices, consulServicesToServices(allConsulServices)...)
 	}
 	return allServices, nil
+}
+
+func dcAwareQueriesForAllDcs(agent *consulapi.Client) ([]*consulapi.QueryOptions, error) {
+	datacenters, err := agent.Catalog().Datacenters()
+	if err != nil {
+		return nil, err
+	}
+
+	var queries []*consulapi.QueryOptions
+	for _, dc := range datacenters {
+		queries = append(queries, &consulapi.QueryOptions{
+			Datacenter: dc,
+		})
+	}
+
+	return queries, nil
 }
 
 func (c *Consul) GetAllServices() ([]*service.Service, error) {
@@ -83,16 +94,13 @@ func (c *Consul) GetAllServices() ([]*service.Service, error) {
 }
 
 func (c *Consul) getAllServices(agent *consulapi.Client) ([]*service.Service, error) {
-	datacenters, err := agent.Catalog().Datacenters()
+	dcAwareQueries, err := dcAwareQueriesForAllDcs(agent)
 	if err != nil {
 		return nil, err
 	}
 	var allInstances []*service.Service
 
-	for _, dc := range datacenters {
-		dcAwareQuery := &consulapi.QueryOptions{
-			Datacenter: dc,
-		}
+	for _, dcAwareQuery := range dcAwareQueries {
 		consulServices, _, err := agent.Catalog().Services(dcAwareQuery)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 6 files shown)


consul/consul.go
```diff
@@ -169,25 +183,53 @@ func (c *Consul) register(service *consulapi.AgentServiceRegistration) error {
...
+			deregisterErrors = append(deregisterErrors, deregisterErr)
 		}
 	}
-	return nil, errors.New(fmt.Sprintf("Couldn't find service matching task id %s", searchedTaskId))
+
+	return utils.MergeErrorsOrNil(deregisterErrors, fmt.Sprintf("deregistering by task %s", taskId))
+}
+
+func (c *Consul) findServicesByTaskId(searchedTaskId apps.TaskId) ([]*service.Service, error) {
+	return c.getServicesUsingProviderWithRetriesOnAgentFailure(func(agent *consulapi.Client) ([]*service.Service, error) {
```
janisz: extract this function

consul/consul.go
```diff
@@ -57,42 +57,50 @@ func (c *Consul) getServicesUsingProviderWithRetriesOnAgentFailure(provide Servi
...
 		}
-		for _, consulService := range allConsulServices {
-			allServices = append(allServices, consulServiceToService(consulService))
-		}
+		allServices = append(allServices, consulServicesToServices(allConsulServices)...)
 	}
 	return allServices, nil
 }
 
+func dcAwareQueriesForAllDcs(agent *consulapi.Client) ([]*consulapi.QueryOptions, error) {
```
janisz: `dcAwareQueriesForAllDcs` → `dcAwareQueriesForAllDCs`
> [Words in names that are initialisms or acronyms (e.g. "URL" or "NATO") have a consistent case. ](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

consul/consul.go
```diff
@@ -169,25 +183,53 @@ func (c *Consul) register(service *consulapi.AgentServiceRegistration) error {
 }
 
 func (c *Consul) DeregisterByTask(taskId apps.TaskId, agentAddress string) error {
-	service, err := c.findServiceByTaskId(taskId)
+	services, err := c.findServicesByTaskId(taskId)
 	if err != nil {
 		return err
+	} else if len(services) == 0 {
+		return errors.New(fmt.Sprintf("Couldn't find any service matching task id %s", taskId))
```
janisz: Use [`fmt.Errorf`](https://golang.org/pkg/fmt/#Errorf)

consul/consul.go
```diff
@@ -169,25 +183,53 @@ func (c *Consul) register(service *consulapi.AgentServiceRegistration) error {
...
+
+		var allFound []*service.Service
+		searchedTag := service.MarathonTaskTag(searchedTaskId)
+		for _, dcAwareQuery := range dcAwareQueries {
+			consulServices, _, err := agent.Catalog().Services(dcAwareQuery)
+			if err != nil {
+				return nil, err
+			}
+			for consulService, tags := range consulServices {
+				if contains(tags, searchedTag) {
```
janisz: if this is true we can jump out of the `dcAwareQuereies` loop since all services of one Task must be in one DC

consul/consul.go
```diff
@@ -169,25 +183,53 @@ func (c *Consul) register(service *consulapi.AgentServiceRegistration) error {
...
+		dcAwareQueries, err := dcAwareQueriesForAllDcs(agent)
+		if err != nil {
+			return nil, err
+		}
+
+		var allFound []*service.Service
+		searchedTag := service.MarathonTaskTag(searchedTaskId)
+		for _, dcAwareQuery := range dcAwareQueries {
+			consulServices, _, err := agent.Catalog().Services(dcAwareQuery)
+			if err != nil {
```
janisz: What if error occurs but interesting Services are in the other DC? Is it tested?

consul/consul.go
```diff
@@ -169,25 +183,53 @@ func (c *Consul) register(service *consulapi.AgentServiceRegistration) error {
...
-		return nil, err
-	}
+func (c *Consul) deregisterMultipleServices(services []*service.Service, taskId apps.TaskId) error {
+	var deregisterErrors []error
 	for _, s := range services {
-		taskId, err := s.TaskId()
-		if err == nil && taskId == searchedTaskId {
-			return s, nil
+		deregisterErr := c.Deregister(s)
+		if deregisterErr != nil {
```
janisz: Can't find test where error occures

utils/errors.go
```diff
@@ -0,0 +1,15 @@
+package utils
+
+import "fmt"
+
+func MergeErrorsOrNil(errors []error, description string) error {
```
janisz: Nice extraction of common function :+1:

consul/consul.go
pbetkier: It would retry with a different agent. That's a general problem with the current flow also in `GetAllServices` and `GetServices` – it's either success in all DCs or nothing, with a couple of retries. We can think of a better flow for all these 3 queries.

consul/consul.go
pbetkier: Yeah ;/ I had no idea how to test that in the current testing infrastructure. Any ideas?

utils/errors_test.go
```diff
@@ -0,0 +1,26 @@
+package utils
```
medzin: Any chance for more meaningful name? :)

utils/errors.go
medzin: Why not create custom type errors that will implement error interface and do all that logic there.

PR #106, opened 2016-08-25
Migrate to marathon task tag

pbetkier:
Adds a script for migrating registered services to the state expected by #99, i.e. containing marathon-task tag with related marathon task ID.

This script is expected to be executed on the consul cluster with services managed by marathon-consul, in all DCs separately, before updating an existing marathon-consul instance. If consul services are not migrated before update, services managed by marathon-consul will get deregistered and registered again according to the new format during sync at first startup following the update. This will cause ~10 seconds service unavailability in the discovery before service healthchecks become passing.

PR #107, opened 2016-08-25
Reregistering on missing marathon-task tag during sync

pbetkier:
When doing sync and coming across a service without marathon-task tag we used to skip that service and leave it in the registry. This change causes the service to get deregistered, as sync's subsequent registering phase should reregister such service properly.

The service will be unavailable for a couple of seconds in this process – before its new healthchecks become passing. If that's a problem, it's advised to add marathon-task tag to all registered services before starting updated marathon-consul, e.g. using migration script from #106.


janisz:
LGTM

PR #110, opened 2016-08-26
Fix travis build

janisz:
It looks like after changing Travis to run `go test ./...` it fails often. This will revert to previous setting when all tests packages were running separately until we find solution for running integration tests.


pbetkier:
👍

PR #112, opened 2016-08-29
Exclude `go-vet` and `go-test` from `goxc`

janisz:
There is an [goxc issue](https://github.com/laher/goxc/issues/90) that
say there won't be workarounds for missing vendoring support in `go vet`
and `go test`. To support building with goxc with vendoring we need to
disable this two tasks.
Related to #101


pbetkier:
👍

PR #114, opened 2016-09-02
Multiple registrations

pbetkier:
This PR enabled registering multiple service instances for a given marathon app, e.g. with different ports, names or tags. These additional registration details can be provided in `portDefinitions` part of the marathon app configuration. Previous configuration method, via app-level `labels`, is still supported and that has not changed.

I'm going to commit changes in README on Monday ;)


apps/app.go
```diff
@@ -49,8 +58,20 @@ func (app *App) IsConsulApp() bool {
 	return ok
 }
 
-func (app *App) ConsulName() string {
-	if value, ok := app.Labels[MARATHON_CONSUL_LABEL]; ok && !isSpecialConsulNameValue(value) {
+func (app *App) HasSameConsulNamesAs(other *App) bool {
+	return reflect.DeepEqual(rawConsulNames(app), rawConsulNames(other))
```
janisz: I prefer to not use `reflect`. How do you guarantee `rawConsulNames` for two different structs will return names in same order.

apps/app.go
```diff
@@ -20,6 +24,10 @@ type HealthCheck struct {
 	}
 }
 
+type PortDefinition struct {
+	Labels map[string]string `json:"labels"`
```
janisz: No `port` field?

apps/app.go
```diff
@@ -73,3 +94,77 @@ func ParseApp(jsonBlob []byte) (*App, error) {
...
+	PortIndex int
+	Labels    map[string]string
+}
+
+func (app *App) findConsulPortDefinitions() []IndexedPortDefinition {
+	var definitions []IndexedPortDefinition
+	for i, d := range app.PortDefinitions {
+		if _, ok := d.Labels[MARATHON_CONSUL_LABEL]; ok {
+			definitions = append(definitions, IndexedPortDefinition{
+				PortIndex: i,
```
janisz: Why you assume port definitions are sorted by port index?

apps/app_test.go
```diff
@@ -119,3 +119,215 @@ func TestAppId_String(t *testing.T) {
 	t.Parallel()
 	assert.Equal(t, "appId", AppId("appId").String())
 }
+
+func TestRegistrationIntent_NameWithSeparator(t *testing.T) {
```
janisz: Think about using [TableDrivenTests](https://github.com/golang/go/wiki/TableDrivenTests) or

janisz: Think about using [TableDrivenTests](https://github.com/golang/go/wiki/TableDrivenTests)

apps/app.go
```diff
@@ -49,8 +58,20 @@ func (app *App) IsConsulApp() bool {
 	return ok
 }
 
-func (app *App) ConsulName() string {
-	if value, ok := app.Labels[MARATHON_CONSUL_LABEL]; ok && !isSpecialConsulNameValue(value) {
+func (app *App) HasSameConsulNamesAs(other *App) bool {
```
janisz: Missing unit test for this function.

apps/app.go
pbetkier: Well, the implementation is deterministic so if configurations are the same then names will also be.

apps/app.go
pbetkier: In `portDefinitions` you can either provide a port explicitly or provide zero in which case the port will be randomly picked. I've added support for the latter only. I guess you're right that we should support both ways.

wendigo: Guys, it's saturday! Get rest :)

janisz: Open source never sleeps.

wendigo: ✊

apps/app.go
janisz: Got it! I'll prefer to replace reflect with simple loop and do not rely on `reflect` package but I don't insist.

consul/consul_test.go
```diff
@@ -385,6 +386,66 @@ func TestRegisterServices_CustomServiceName(t *testing.T) {
...
+	services, _ := consul.GetAllServices()
+	sort.Sort(&serviceSorter{
+		services: services,
+		by:       func(s1, s2 *service.Service) bool { return s1.Name < s2.Name },
+	})
+
+	// then
+	assert.Len(t, services, 2)
+	assert.Equal(t, "first-name", services[0].Name)
+	assert.Equal(t, []string{"marathon", "common-tag", "first-tag", "marathon-task:serviceA.0"}, services[0].Tags)
```
medzin: Why not just use [assert.Contains][https://godoc.org/github.com/stretchr/testify/assert#Contains] here?

pbetkier: I would have to call assert.Contains 4 times and check the lengths to make sure it contains these tags and not other...

janisz: Add dedicated method to compare slices so it will be clear

pbetkier: How would it be different from checking if two slices are equal the way I did?

apps/app.go
pbetkier: Added missing unit test.

consul/consul_test.go
janisz: No difference but it will be clear why sorting is required.

pbetkier:
I've added support for explicit ports, reworked checking if two apps have same names, added table-driven tests. What do you think now?

I'm working on the README at the moment, going to push that soon.

consul/consul_test.go
pbetkier: Oh, we're speaking of 2 different things. Sorting is not for this line with tags, it's for picking the right services for assertions. I'll think how to rework it to make it more clear.

janisz: How about comparing result of `consul.GetAllServices()` with expected list of services?

medzin: Maybe GetAllServices should return sorted list, or it doesn't matter in not test code?

README.md
```diff
@@ -215,14 +217,76 @@ Endpoint  | Description
 `/health` | healthcheck - returns `OK`
 `/events` | event sink - returns `OK` if all keys are set in an event, error message otherwise
 
-### Known limitations
+## Advanced usage
```
janisz: How about naming this section "Multiple ports support" since it describes how to do it.

dankraw: Or maybe we should add a sub section called 
```markdown
### Multiple ports support
```

README.md
```diff
@@ -215,14 +217,76 @@ Endpoint  | Description
...
+via Marathon's `portDefinitions`:
 
-* In Marathon when a deployment changing the application's service name (by changing its `labels`) is being stopped, it changes app's configuration anyway.
-  This means we loose the link between the app and the services registered with the old name in Consul.
-  Later on, if another deployment takes place, new services are registered with a new name, the old ones are not being deregistered though.
-  A scheduled sync is required to wipe them out.
+```
+  "id": "my-new-app",
+  "labels": {
+    "consul": "true",
```
dankraw: Isn't `true` a *deprecated* value for this label?
It should be left blank if consul service name is going to be equal to marathon app name. 
Or we should remove the deprecation note if we can't get rid of this feature. :)

README.md
```diff
@@ -215,14 +217,76 @@ Endpoint  | Description
...
+      "labels": {
+        "consul": "my-app-custom-name"
+      }
+    },
+    {
+      "port": 0,
+      "protocol": "tcp",
+      "labels": {
+        "consul": "my-app-other-name",
+        "specific-tag": "tag
```
dankraw: closing `"` missing

README.md
pbetkier: Oh, ok. I can change it to empty. @janisz what do you think?

README.md
pbetkier: damn, you got me :) ok, I confess, I wrote these examples by hand :(

apps/app_test.go
```diff
@@ -119,3 +119,375 @@ func TestAppId_String(t *testing.T) {
...
+		Labels: map[string]string{"consul": "true"},
+		PortDefinitions: []PortDefinition{
+			PortDefinition{
+				Labels: map[string]string{"consul": "app-name"},
+			},
+		},
+	}
+
+	// expect
+	assert.True(t, app.HasSameConsulNamesAs(other))
```
dankraw: the opposite way assertion `other.HasSameConsulNamesAs(app))` would be welcome :)

consul/consul.go
```diff
@@ -257,42 +269,28 @@ func (c *Consul) deregister(toDeregister *service.Service) error {
...
-	}
-	return serviceName
-}
-
-func (c *Consul) marathonAppNameToConsulServiceName(name string) string {
-	return strings.Replace(strings.Trim(strings.TrimSpace(name), "/"), "/", c.config.ConsulNameSeparator, -1)
-}
-
-func (c *Consul) marathonTaskToConsulService(task *apps.Task, app *apps.App) (*consulapi.AgentServiceRegistration, error) {
+func (c *Consul) marathonTaskToConsulService(task *apps.Task, app *apps.App) ([]*consulapi.AgentServiceRegistration, error) {
```
dankraw: renaming it to `marathonTaskToConsulServices` would be nice :)

web/event_handler.go
```diff
@@ -218,36 +218,13 @@ func (fh *eventHandler) handleDeploymentStepSuccess(body []byte) error {
...
-
-	if len(services) == 0 {
-		log.WithField("AppId", app.ID).WithField("ServiceName", serviceName).Info("No matching Consul services found")
-		return errors
-	}
+	log.WithField("AppId", app.ID).Info("Deregistering all services")
 
-	for _, service := range services {
-		taskId, err := service.TaskId()
+	for _, task := range app.Tasks {
```
dankraw: Are we sure the `app.Tasks` will be available? Like in this case from `event_handler.go`:

```go
	for _, app := range deploymentEvent.StoppedConsulApps() {
		for _, err = range fh.deregisterAllAppServices(app) {
			errors = append(errors, err)
		}
	}
```

README.md
dankraw: still fine :)

README.md
janisz: I opt for using empty name. I think it's time to cleanup `consul: true` logic but not in this PR.

web/event_handler.go
pbetkier: Oh I see, so that's why this logic was so complicated? Test missing! :)

dankraw: `TestWebHandler_HandleDeploymentInfoWithStopApplicationAction` looks like a test that should fail here, but it might be wrong (generated json request body).

pbetkier: Ok I fixed the test and the implementation, check this out @dankraw @janisz

janisz:
LGTM :+1:

PR #115, opened 2016-09-07
Sync support for multi-registrations

pbetkier:
In multi-registration scenario, some, but not all, service registrations may fail. Sync should check if all registrations are in Consul and register the missing ones.

Change base to *develop* once #114 is merged.

### DIFF
--- a/apps/app.go
+++ b/apps/app.go
@@ -119,6 +119,19 @@ type RegistrationIntent struct {
 	Tags []string
 }
 
+func (app *App) RegistrationIntentsNumber() int {
+	if !app.IsConsulApp() {
+		return 0
+	}
+
+	definitions := app.findConsulPortDefinitions()
+	if len(definitions) == 0 {
+		return 1
+	}
+
+	return len(definitions)
+}
+
 func (app *App) RegistrationIntents(task *Task, nameSeparator string) []*RegistrationIntent {
 	commonTags := labelsToTags(app.Labels)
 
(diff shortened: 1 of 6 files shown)


sync/sync.go
```diff
@@ -143,14 +143,19 @@ func (s *Sync) deregisterConsulServicesNotFoundInMarathon(marathonApps []*apps.A
...
+	registrationsUnderTaskIds := s.taskIdsInConsulServices(services)
 	for _, app := range marathonApps {
 		if !app.IsConsulApp() {
 			log.WithField("Id", app.ID).Debug("Not a Consul app, skipping registration")
 			continue
 		}
+		expectedRegistrations := app.RegistrationIntentsNumber()
 		for _, task := range app.Tasks {
-			if _, ok := registeredTaskIds[task.ID]; !ok {
+			if registrations := registrationsUnderTaskIds[task.ID]; registrations != expectedRegistrations {
```
janisz: This function is not clear. Extract fors body into another fuction

apps/app.go
```diff
@@ -119,6 +119,19 @@ type RegistrationIntent struct {
 	Tags []string
 }
 
+func (app *App) RegistrationIntentsNumber() int {
```
pbetkier: I should probably add a unit test for this.

sync/sync.go
```diff
@@ -166,12 +171,11 @@ func (s *Sync) registerAppTasksNotFoundInConsul(marathonApps []*apps.App, servic
 	}
 }
 
-func (s *Sync) taskIdsInConsulServices(services []*service.Service) map[apps.TaskId]struct{} {
-	servicesSet := make(map[apps.TaskId]struct{})
-	var exists struct{}
+func (s *Sync) taskIdsInConsulServices(services []*service.Service) map[apps.TaskId]int {
+	servicesSet := make(map[apps.TaskId]int)
```
janisz: It's not a set anymore.

PR #117, opened 2016-09-08
Add linter and fix issues

janisz:
Added `gometalinter` to as a Travis build step. Fixed couple of issues (mostly naming). We can't introduce `golint` until we add comment to public function/struct/type/variable/constant.


dankraw:
👍

PR #120, opened 2016-09-12
Blacklisted Healthcheck types

### LINKED ISSUE #118: Disabling command healthchecks registration
Currently marathon-consul migrates all healthchecks from marathon/mesos to consul. I would like to have a possibility to disable migration for healthchecks type of command.

dankraw:
Introducing `consul-ignored-healthchecks` configuration property which allows one to ignore specified types of Marathon health checks during service registration in Consul.

Fixes #118

### DIFF
--- a/consul/consul.go
+++ b/consul/consul.go
@@ -305,7 +305,13 @@ func (c *Consul) serviceID(task *apps.Task, name string, port int) string {
 func (c *Consul) marathonToConsulChecks(task *apps.Task, healthChecks []apps.HealthCheck, serviceAddress string) consulapi.AgentServiceChecks {
 	var checks = make(consulapi.AgentServiceChecks, 0, len(healthChecks))
 
+	ignoredHealthCheckTypes := c.getIgnoredHealthCheckTypes()
 	for _, check := range healthChecks {
+		if contains(ignoredHealthCheckTypes, check.Protocol) {
+			log.WithField("Id", task.AppID.String()).WithField("Address", serviceAddress).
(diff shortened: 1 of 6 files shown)


consul/consul.go
```diff
@@ -349,6 +349,31 @@ func (c *Consul) marathonToConsulChecks(task *apps.Task, healthChecks []apps.Hea
...
+	for _, check := range healthChecks {
+		if !contains(ignoredTypes, check.Protocol) {
+			filteredChecks = append(filteredChecks, check)
+		}
+	}
+	return filteredChecks
+}
+
+func (c *Consul) getIgnoredHealthCheckTypes() []string {
+	ignoredTypes := make([]string, 0)
```
janisz: How about something like this:
```
ignoredTypes := range strings.Split(strings.ToUpper(c.config.IgnoredHealthChecks), ",")
```

druminski:
:+1:

consul/consul.go
dankraw: OK, I've fixed it so `ToUpper` is called once, but still I'd like to check if the input is valid (trim, remove empty values):
```go
ignoredTypes := make([]string, 0)
	for _, ignoredType := range strings.Split(strings.ToUpper(c.config.IgnoredHealthChecks), ",") {
		var ignoredType = strings.TrimSpace(ignoredType)
		if ignoredType != "" {
			ignoredTypes = append(ignoredTypes, ignoredType)
		}
	}
	return ignoredTypes
```
By default it returns empty splice.

janisz:
LGTM :+1:

PR #124, opened 2016-09-29
Skip tasks with excess registrations when syncing

pbetkier:
Explicitly handle a rare scenario during sync in which a task has more registrations than it should have according to its application configuration.

This may happen when application configuration is changed (now defining less registrations) and then the deployment this caused is rolled back – the new application configuration still stays in Marathon. Or we end up in some weird state. Since we don't know what to do in such scenario – we don't know which service to deregister – I suggest skipping such task with a log message.

We could also deregister all registrations and register the task again, but this seemed like a bad idea to me – the deployment with this new configuration was rolled back after all!

### DIFF
--- a/sync/sync_test.go
+++ b/sync/sync_test.go
@@ -349,6 +349,36 @@ func TestSync_ShouldRegisterMissingRegistrationInMultiregistrationScenario(t *te
 	assert.Len(t, services, 2)
 }
 
+/*
+This may happen if an application configuration is changed and then the deployment this triggered is rolled back.
+After rollback, the new configuration still stays with the application and there's no way to access the original configuration
+that was used to start the currently running tasks. In such case, it's possible that a given task has more registrations
+than it's now expected from the new application configuration. In order to be safe we don't want to deregister anything,
+let someone make the deployment explicitly.
+ */
+func TestSync_SkipServiceHavingMoreRegistrationsThanExpectedInMultiregistrationScenario(t *testing.T) {
+	t.Parallel()
+	// given
+	app := ConsulAppMultipleRegistrations("/test/app", 1, 2)
+	marathon := marathon.MarathonerStubForApps(app)
+	consul := consul.NewConsulStub()
+
+	consul.Register(&app.Tasks[0], app)
+	services, _ := consul.GetAllServices()
+	assert.Len(t, services, 2)
+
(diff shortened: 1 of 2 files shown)


sync/sync_test.go
```diff
@@ -349,6 +349,36 @@ func TestSync_ShouldRegisterMissingRegistrationInMultiregistrationScenario(t *te
 	assert.Len(t, services, 2)
 }
 
+/*
+This may happen if an application configuration is changed and then the deployment this triggered is rolled back.
+After rollback, the new configuration still stays with the application and there's no way to access the original configuration
```
janisz: Do you mean rollback with force (aka stop deployment)  or regular rollback?

pbetkier: Stop deployment.

janisz: Then, it should be explicitly mentioned that this is special case for deployment cancelation or add example request that trigger this state 

    DELETE /v2/deployments/177b7556-1287-4e09-8432-3d862981a987?force=true

sync/sync_test.go
```diff
@@ -350,7 +350,7 @@ func TestSync_ShouldRegisterMissingRegistrationInMultiregistrationScenario(t *te
 }
 
 /*
-This may happen if an application configuration is changed and then the deployment this triggered is rolled back.
+This may happen if an application configuration is changed and then the deployment this triggered is rolled back (deployment stopped).
```
kamilchm: There's no such thing as **stopping deployment** in current releases of Marathon. You can `cancel`/`delete` deployment **with** or **without** a new **rollback deployment** https://github.com/mesosphere/marathon/blob/master/docs/docs/rest-api/mesosphere/marathon/api/v2/DeploymentsResource_cancel.md

sync/sync_test.go
pbetkier: Actually this may happen in many ways. E.g. you change configuration, deployment starts, but for some reason the application cannot start so deployment spins endlessly. Configuration gets updated though, so during sync the existing tasks refer to this new configuration. That's even before you click deployment stop. I'm going to update this documentation comment to reflect this.

pbetkier: @janisz how about now?

janisz: LGTM

Issue #122, opened 2016-09-21
Set initial consul health check to passing


janisz:
Consul allows setting initial health check value. Since we know that application is working (we register only when Marathon health check is passing) we can minimize time when service is unavailable in consul by setting health check to passing when registering service. 

https://github.com/hashicorp/consul/issues/2314#issuecomment-248630431

PR #125, opened 2016-09-30
Register first, deregister second in sync process

pbetkier:
Switching registering and deregistering places for smooth migration from 0.4.x to 1.0.0, without having to migrate using this script #106. We want to register services in the new format before removing the old entries to avoid unavailability. Shouldn't make a difference for the synchronization process. @janisz do you see any potential problem with that?

### DIFF
--- a/sync/sync.go
+++ b/sync/sync.go
@@ -76,8 +76,8 @@ func (s *Sync) syncServices() error {
 		return fmt.Errorf("Can't get Consul services: %v", err)
 	}
 
-	s.deregisterConsulServicesNotFoundInMarathon(apps, services)
 	s.registerAppTasksNotFoundInConsul(apps, services)
+	s.deregisterConsulServicesNotFoundInMarathon(apps, services)
 
 	log.Info("Syncing services finished")
 	return nil


janisz:
In #122 we minimize gap between registering service and making it healthy. I don't see any drwabacks of reversing order of registering and deregistering. I think migration script is still the safest way of upgrading to 1.0 but with this PR accidental version changes will not cause huge outage.

pbetkier:
Turns out migrating by just adding marathon-task tag, without changing the service id in consul, doesn't work as smooth as we expected.

From time to time, for every task, we get an event from marathon specifying that its healthcheck is now passing. We interpret this in marathon-consul as having to register this task. Since tasks are deterministically mapped to service ids this usually doesn't have any effect – just attempting to reregister service under the same id. But since we migrated by just adding a tag without changing the service id to match the new scheme, the service gets registered for the second time and we end up with two entries in consul per service. This stays until such service is redeployed and the old entry gets finally removed from consul.

I would go with sync as the suggested way to migrate to 1.x, now that we are able to register services in consul as healthy, and remove the migration script altogether.

janisz:
Good catch! Maybe we should remove migration script.

PR #129, opened 2016-10-10
Simplify stopping and renaming applications

pbetkier:
This PR removes handling of StopApplication and RestartApplication events, which caused problems in multi-registration scenario and required some rework. It seems we don't need them anymore though, TASK_KILLED events that we handle seem to cover all the cases.


pbetkier:
Just for the record, the issue is that in multi-registration scenario an application maps to multiple names, so after a rename it's possible that some name from the previous configuration still exists in the new one. If we deregister all services matching names from the just stopped / renamed configuration then in such case we also deregister newly registered services from tasks of the new configuration!

PR #132, opened 2016-10-12
Fixes #113 | Use fpm to create packages

### LINKED ISSUE #113: Deb package should not replace configs
When you install marathon-consul from deb it will overwrite `/etc/marathon-consul.d/config.json`. Instead user should be asked what should happen.

https://www.debian.org/doc/debian-policy/ch-files.html#s-config-files

janisz:
With this PR we are switching from goxc to fpm to create Debian/Ubuntu packages.


pbetkier:
👍

PR #133, opened 2016-10-13
Fix tests on tip (golang 1.8)

wendigo:
In Go 1.8 json Marshaler/Unmarshaler gives more detailed errors and one of the tests depends on the exact error message.

I fixed the failing test - I'm not really sure if approach is right so I will leave it here to discuss it further

### DIFF
--- a/web/event_handler_test.go
+++ b/web/event_handler_test.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"runtime"
+	"strings"
 	"testing"
 
 	"github.com/allegro/marathon-consul/apps"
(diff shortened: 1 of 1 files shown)


web/event_handler_test.go
```diff
@@ -111,7 +113,12 @@ func TestWebHandler_NotHandleInvalidEventType(t *testing.T) {
 
 	// then
 	assert.Equal(t, 400, recorder.Code)
-	assert.Equal(t, "json: cannot unmarshal array into Go value of type string\n", recorder.Body.String())
+
+	if runtime.Version() < "go1.8" && !strings.HasPrefix(runtime.Version(), "devel") {
+		assert.Equal(t, "json: cannot unmarshal array into Go value of type string\n", recorder.Body.String())
```
janisz: Maybe just check is error is not nil.

PR #140, opened 2016-11-16
Drop not supported events

janisz:


janisz:
Results of synthetic benchmark with 400MB of real events.

```
$ git co master && go test -bench=BenchmarkWebHandler -benchtime=10s ./web -test.benchmem  > old.txt && git co fixes/\#139/drop_not_supported_events  && go test -bench=BenchmarkWebHandler -benchtime=10s ./web -test.benchmem  > new.txt && benchcmp old.txt new.txt

benchmark                 old ns/op     new ns/op     delta
BenchmarkWebHandler-4     808854977     690828083     -14.59%

benchmark                 old allocs     new allocs     delta
BenchmarkWebHandler-4     2628254        1900092        -27.71%

benchmark                 old bytes     new bytes     delta
BenchmarkWebHandler-4     374338713     276159748     -26.23%

```

<details>
Benchmark does only test change in handling events. Does not include speed up with less events pushed to event queue and processing them. More details could be find here https://gist.github.com/janisz/9db4d50fc09f2eba81781af8dbc26a03
```go
package web

import (
	"bytes"
	"net/http/httptest"
	"testing"
	"os"
	"bufio"
	"net/http"
	"io"
)

func BenchmarkWebHandler(b *testing.B) {

	// given
	queue := make(chan event)
	handler := newWebHandler(queue)
	f, err := os.Open("events.jsons")
	if err != nil {
		b.Fatal(err)
	}
	eventsReader := bufio.NewReader(f)

	go func() {
		for {
			<-queue
		}
	}()

	events := make([]*http.Request, 0, 3634)
	for {
		event, _, err := eventsReader.ReadLine()

		if err == io.EOF {
			break
		}
		if err != nil {
			b.Fatal(err)
		}
		req, _ := http.NewRequest("POST", "/events", bytes.NewBuffer(event))
		events = append(events, req)
	}

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		for _, event := range events {
			handler.Handle(httptest.NewRecorder(), event)
		}
	}
}
```
</details>

kamilchm:
It would be good to replace it with implementation of https://github.com/mesosphere/marathon/issues/4637 and use in the future.

PR #142, opened 2016-11-18
Parse event timestamp and add more events metrics

janisz:


metrics/metrics.go
```diff
@@ -24,6 +24,11 @@ func Mark(name string) {
 	meter.Mark(1)
 }
 
+func Histogram(name string, value int64) {
+	histogram := metrics.GetOrRegisterHistogram(name, metrics.DefaultRegistry, metrics.NewExpDecaySample(1028, 0.015))
```
adamdubiel: should we remove histograms in favor of gauge?

PR #144, opened 2016-11-25
Implement SSE Client

janisz:
@wendigo @medzin Can you take a look at this and give some early feedback?


sse_client/client.go
```diff
@@ -0,0 +1,74 @@
...
+
+type client struct {
+	request   *http.Request
+	client    *http.Client
+	close     func()
+	onMessage func(Event)
+	onError   func(error)
+}
+
+func NewEventSource(url *url.URL, httpClient *http.Client, onMessage func(Event), onError func(error)) (*client, error) {
```
medzin: You probably want to return `EventSource` not `*client`.

sse_client/event_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+	// when
+	_, err := parseEvent(events)
+
+	// then
+	assert.EqualError(t, err, "Unexpected EOF")
+}
+
+func TestParseEvent(t *testing.T) {
+	testCases := []struct {
```
janisz: JSON does not support multi line strings. How this should be done? This Test cases are really small. What I will achieve with moving them to file?

sse_client/client.go
janisz: Fixed

sse_client/client.go
```diff
@@ -0,0 +1,76 @@
...
+	//Open starts reading form source
+	Open()
+	//Close connection
+	Close()
+}
+
+type client struct {
+	request   *http.Request
+	client    *http.Client
+	close     func()
```
medzin: Use a `context.CancelFunc` here instead of the raw `func()` type.

janisz: Done

sse_client/client.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"bufio"
+	"context"
+	"io"
+	"net/http"
+	"net/url"
+)
+
+type EventSource interface {
+	//Open starts reading form source
```
wendigo: I think that Open and Close should return error

sse_client/client.go
```diff
@@ -0,0 +1,76 @@
...
+		onMessage: onMessage,
+		onError:   onError,
+		close:     cancel,
+	}
+}
+
+func (c client) Open() {
+	response, err := c.client.Do(c.request)
+	if err != nil {
+		c.onError(err)
```
wendigo: IMO: should return error here

sse_client/client.go
```diff
@@ -0,0 +1,76 @@
...
+		return
+	}
+
+	go func() {
+		defer response.Body.Close()
```
wendigo: Should close() response.Body in Close() method I guess.

sse_client/client.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func (c client) Open() {
+	response, err := c.client.Do(c.request)
+	if err != nil {
+		c.onError(err)
+		return
+	}
+
+	go func() {
```
wendigo: What about cancelation? Maybe add Context to Open() ?

sse_client/client.go
```diff
@@ -0,0 +1,76 @@
...
+		defer response.Body.Close()
+		//TODO: Consider using bufio.Scanner
+		reader := bufio.NewReader(response.Body)
+		for {
+			e, err := parseEvent(reader)
+			if err != nil {
```
wendigo: maybe instead of if use switch statement (if it's possible):

```
switch (err) {
    case nil:
    case io.EOF:
        c.onMessage(e)
    default:
      c.onError(err)
}
```

sse_client/client.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func NewEventSource(url *url.URL, httpClient *http.Client, onMessage func(Event), onError func(error)) EventSource {
+	req := &http.Request{
+		Method:     http.MethodGet,
+		URL:        url,
+		Proto:      "HTTP/1.1",
+		ProtoMajor: 1,
+		ProtoMinor: 1,
+		Header:     make(http.Header),
```
wendigo: http.Header{}

sse_client/event.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// parseLine is implementation of event stream interpretation w3c standard. Is designed to be run multiple time.
+// `line` argument should be line readed from buffer without new line character.
+// Returns true when event is ready for dispatch
+// See: https://www.w3.org/TR/eventsource/#event-stream-interpretation
+// Reference implementation: https://github.com/WebKit/webkit/blob/9f191f/Source/WebCore/page/EventSource.cpp#L272-L363
+func (e *Event) parseLine(line []byte) bool {
+
+	line = bytes.TrimSuffix(line, []byte{'\n'})
```
wendigo: line = bytes.TrimRight(line, "\n\r") ?

sse_client/event.go
```diff
@@ -0,0 +1,119 @@
...
+	//println(string(line))
+
+	//If the line is empty (a blank line)
+	if len(line) == 0 {
+		//Dispatch the event, as defined below.
+		return !e.isEmpty()
+	}
+
+	//If the line starts with a U+003A COLON character (:)
+	if bytes.HasPrefix(line, []byte{':'}) {
```
wendigo: instead of allocating it every time introduce it as var. same applies to every other occurance

sse_client/event.go
```diff
@@ -0,0 +1,119 @@
+package sse_client
```
wendigo: Maybe use https://github.com/donovanhide/eventsource instead?

sse_client/client.go
janisz: I was thinking about this. It was implemented this way at the beginning, but then you could require user to handle one error twice.

sse_client/client.go
janisz: When you call `Close()` request will be cancelled and loop in this function will be break on `EOF` so it close will be called.

sse_client/event.go
janisz: I've must miss it when I search for SSE implementation. Unfortunately I think we need to stick with custom implementation. We need to detect if we are connected to master. And the only way to do it is to check response headers.

wendigo: You can still use Decoder (https://github.com/donovanhide/eventsource/blob/master/decoder.go) instead of implementing it yourself.

wendigo: Why do we need to detect if we are connected to master?

janisz: Becouse we have multiple instances of marathon-consul but we want only one of them to register services. When leader abdicate it connection will be closed and we will once again connect and check if we are connected to master.

Any idea how to do it better?

sse_client/event.go
janisz: Are you sure this leads to allocation?

PR #147, opened 2016-12-01
Fixes #56 | Humman-readable interval in config

### LINKED ISSUE #56: Human-readable time durations in config files
It would be nice if we had human-readable time durations in config files instead of the nanosecond integers, for example:

``` json
"Sync": {
  "Interval": 900000000000
}
```

would become

``` json
"Sync": {
  "Interval": "15m"
}
```

janisz:
Currently we allow user to specify intervals/timeouts in
human readable format only in command line flags. This commit
introduce allowign user to specify duration as inteager nanoseconds or
as formated string. String should have format supported by
time.ParseDuration.

### DIFF
--- a/time/time.go
+++ b/time/time.go
@@ -1,0 +1,51 @@
+package time
+
+import (
+	"encoding/json"
+	"strconv"
+	"strings"
+	"time"
+)
+
+type Timestamp struct {
+	time.Time
+}
+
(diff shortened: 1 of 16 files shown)


time/time.go
```diff
@@ -0,0 +1,51 @@
+package time
```
medzin: Maybe `javatime` would fit better?

PR #153, opened 2016-12-22
make receiving events nonblocking even when queue is full

adamdubiel:
Also added two metrics:

* discarded (dropped) events when queue is full
* event queue utilization on 0-100 scale

### DIFF
--- a/web/event_handler.go
+++ b/web/event_handler.go
@@ -55,7 +55,11 @@ func (fh *eventHandler) start() chan<- stopEvent {
 			select {
 			case e = <-fh.eventQueue:
 				metrics.Mark(fmt.Sprintf("events.handler.%d", fh.id))
-				metrics.UpdateGauge("events.queue.len", int64(len(fh.eventQueue)))
+
+				ql := int64(len(fh.eventQueue))
+				metrics.UpdateGauge("events.queue.len", ql)
+				metrics.UpdateGauge("events.queue.util", 100*(ql/int64(cap(fh.eventQueue))))
+
 				metrics.UpdateGauge("events.queue.delay_ns", time.Since(e.timestamp).Nanoseconds())
 				metrics.Time("events.processing."+e.eventType, process)
 			case <-quitChan:
--- a/web/event_handler_test.go
+++ b/web/event_handler_test.go
@@ -20,7 +20,7 @@ func TestWebHandler_DropUnknownEventType(t *testing.T) {
 	t.Parallel()
 
 	// given
-	queue := make(chan event)
+	queue := make(chan event, 10)
 	handler := newWebHandler(queue, maxEventSize)
 	stopChan := newEventHandler(0, nil, nil, queue).start()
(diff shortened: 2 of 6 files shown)


web/event_handler.go
```diff
@@ -20,20 +20,22 @@ type event struct {
...
 type eventHandler struct {
-	id              int
-	serviceRegistry service.ServiceRegistry
-	marathon        marathon.Marathoner
-	eventQueue      <-chan event
+	id                 int
+	serviceRegistry    service.ServiceRegistry
+	marathon           marathon.Marathoner
+	eventQueue         <-chan event
+	eventQueueCapacity int
```
janisz: Do we really need this value? Can't just use [cap](https://golang.org/pkg/builtin/#cap)

adamdubiel: oh great, this is what i was looking for and did not find, will change

adamdubiel:
Are build always this flaky?

janisz:
Tests get flaky when they are run in parallel (multiple packages at once) but in Travis we have run them sequentially. 
Unfortunately your code doesn't pass linter https://travis-ci.org/allegro/marathon-consul/builds/186278642#L217

adamdubiel:
Oh, okay, linter this time. But if you see build history, each time different tests were failing (build continues after 1st fail).

adamdubiel:
Found two places that made tests at my localhost (Mac) fail:

* bumped Consul version to 0.7.2, which fixed issue with Consul segfaults at random moments
* fixed race condition detected by `-race` in Marathon stub

I also added 2 new targets to Makefile:

* `format` that runs `go fmt` on `marathon-consul` sources
* `check` that runs `gometalinter` in same config as on travis

web/event_handler.go
```diff
@@ -55,7 +55,11 @@ func (fh *eventHandler) start() chan<- stopEvent {
 			select {
 			case e = <-fh.eventQueue:
 				metrics.Mark(fmt.Sprintf("events.handler.%d", fh.id))
-				metrics.UpdateGauge("events.queue.len", int64(len(fh.eventQueue)))
+
+				ql := int64(len(fh.eventQueue))
```
janisz: Use meaningful name for variable `queueLength`

web/event_handler_test.go
```diff
@@ -20,7 +20,7 @@ func TestWebHandler_DropUnknownEventType(t *testing.T) {
 	t.Parallel()
 
 	// given
-	queue := make(chan event)
+	queue := make(chan event, 10)
```
janisz: Why you use buffered channel?

web/event_handler.go
adamdubiel: Yeah, looks like the whole go world likes to use those silly one or two letter names, just tried to follow ;) Will change

web/event_handler_test.go
adamdubiel: might not make sense anymore, but it helped to pass any tests at the beginning. will change

adamdubiel:
i'm going to close this pr and try to work on tests. they fail randomly both on localhost and travis when using nonblocking channel.

PR #154, opened 2016-12-26
Nonblocking events

adamdubiel:
In addition to adding nonblocking handling of events i added:

* `make format` and `make check` that runs same metalinter command as travis.ci
* bumped Consul version to 0.7.2 to fix random segfaults on Mac
* fixed race condition detected by `go test -race` in `MarathonerStub`
* fixed race condition in `even_hander_test` that caused tests to fail randomly (2nd commit)

Bit more about test fixes. I moved `web_handler` test suite to `web_handler_test` and it tests only `web_handler`, the event queue is stubbed. `event_handler` tests on the other hand no longer start `web_handler`, instead events that should be processed are sent directly to the event queue. This fixes a non-trivial race condition in tests, where success of test depended on wether request will be processed by `web_handler` and sent to the queue before test reaches synchronisation point at blocking send to `event_handler` `stopChan`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,11 @@
+SHELL := /bin/bash
+
 PACKAGES=$(shell go list ./... | grep -v /vendor/)
 TESTARGS?=-race
 CURRENT_DIR = $(shell pwd)
 SOURCEDIR = $(CURRENT_DIR)
 SOURCES := $(shell find $(SOURCEDIR) -name '*.go')
+APP_SOURCES := $(shell find $(SOURCEDIR) -name '*.go' -not -path '$(SOURCEDIR)/vendor/*')
 VERSION=$(shell cat .goxc.json | python -c "import json,sys;obj=json.load(sys.stdin);print obj['PackageVersion'];")
 TEMPDIR := $(shell mktemp -d)
 LD_FLAGS=-ldflags '-w' -ldflags "-X main.VERSION=$(VERSION)"
@@ -26,6 +29,16 @@ docker: build-linux
 test: deps $(SOURCES)
 	PATH=$(CURRENT_DIR)/bin:$(PATH) go test $(PACKAGES) $(TESTARGS)
 	go vet $(PACKAGES)
+
+gometalint-exists:
+	@which gometalinter > /dev/null || \
+	(echo >&2 "gometalinter must be installed on the system. See https://github.com/alecthomas/gometalinter")
+
+check: gometalint-exists $(SOURCES)
+	gometalinter . --deadline  720s --vendor -D dupl -D gotype -D errcheck -D gas -D golint -E gofmt
+
+format:
+	goimports -w -l $(APP_SOURCES)
 
 FPM-exists:
 	@fpm -v || \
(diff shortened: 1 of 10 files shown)


marathon/marathon_stub.go
```diff
@@ -4,46 +4,56 @@ import (
...
-	AppStub      map[apps.AppID]*apps.App
-	TasksStub    map[apps.AppID][]*apps.Task
-	leader       string
-	interactions bool
+	AppsStub       []*apps.App
+	AppStub        map[apps.AppID]*apps.App
+	TasksStub      map[apps.AppID][]*apps.Task
+	leader         string
+	interactionsMu sync.RWMutex
+	interactions   bool
```
janisz: How about changing interactions to int and count them. Then we can use [atomic](https://golang.org/pkg/sync/atomic/) functions on it and we do not need mutex.

web/event_handler_test.go
```diff
@@ -1,144 +1,41 @@
...
-	// when
-	recorder := httptest.NewRecorder()
-	handler.Handle(recorder, req)
-	stopChan <- stopEvent{}
+	awaitChan := newEventHandler(0, stubs.serviceRegistry, stubs.marathon, queue).start()
 
-	// then
-	assertDropped(t, recorder)
+	// stopChan returned by newEventHandler can be used as a synchronization
+	// point to wait until previous event has been processed
```
janisz: This comment should be converted to function comment

web/web_handler.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/allegro/marathon-consul/events"
 	"github.com/allegro/marathon-consul/metrics"
 
+	"errors"
```
janisz: goimports

marathon/marathon_stub.go
wendigo: When using atomic struct containing such field must be properly aligned: https://github.com/golang/go/issues/13868

janisz: I don't think it's our case. We don't target to ARM or x86 and we only need `uint32`

Makefile
```diff
@@ -27,6 +29,13 @@ test: deps $(SOURCES)
 	PATH=$(CURRENT_DIR)/bin:$(PATH) go test $(PACKAGES) $(TESTARGS)
 	go vet $(PACKAGES)
 
+gometalint-exists:
+	@which gometalinter > /dev/null || \
```
adamdubiel: good idea, but i would rather not have makefile installing dependencies when running build like this. instead we can create new target like `install-tools` which would do this job, and error message in `-exists` tasks would point to running `make install-tools`. So i will leave this as it is for now and we can introduce this additional target along with changing travis.yaml to use Make

marathon/marathon_stub.go
adamdubiel: I would rather leave it with mutex for simplicity/readability sake - no need for "smart" code here, this is just test stub (not hot part of code) and from what i see bool is enough. We can add counter when/If we need it.

Makefile
janisz: Agree. You can add it to [`deps`](https://github.com/allegro/marathon-consul/blob/3607a6be4c1bba0843d42dcf62886e4a46f13183/Makefile#L14) target

adamdubiel:
@janisz i added requested changes: split change into commits and `goimports` everywhere

PR #155, opened 2016-12-29
Move all build logic and dependencies installs to Makefile

adamdubiel:
travis.ci configuration uses make commands to be able to replicate CI builds
on local machine

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,26 +3,7 @@ language: go
 go:
   - 1.7
 before_install:
-  - go get golang.org/x/tools/cmd/cover
-  - go get github.com/axw/gocov/gocov
-  - go get github.com/modocache/gover
   - go get github.com/mattn/goveralls
-  - if ! go get github.com/golang/tools/cmd/cover; then go get golang.org/x/tools/cmd/cover; fi
-  - go get github.com/alecthomas/gometalinter
 script:
-  - make deps
-  - gometalinter --install
-  - gometalinter ./... --deadline 720s --vendor -D dupl -D gotype -D errcheck -D gas -D golint -E gofmt
-  - go test -coverprofile=apps.coverprofile ./apps
-  - go test -coverprofile=config.coverprofile ./config
-  - go test -coverprofile=consul.coverprofile ./consul
-  - go test -coverprofile=events.coverprofile ./events
-  - go test -coverprofile=marathon.coverprofile ./marathon
-  - go test -coverprofile=metrics.coverprofile ./metrics
-  - go test -coverprofile=service.coverprofile ./service
-  - go test -coverprofile=sync.coverprofile ./sync
-  - go test -coverprofile=utils.coverprofile ./utils
-  - go test -coverprofile=web.coverprofile ./web
-  - go test -coverprofile=main.coverprofile
-  - $HOME/gopath/bin/gover
-  - $HOME/gopath/bin/goveralls -coverprofile=gover.coverprofile -service travis-ci
+  - make check
+  - goveralls -coverprofile=coverage/gover.coverprofile -service travis-ci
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -3,26 +3,7 @@ language: go
 go:
   - 1.7
 before_install:
-  - go get golang.org/x/tools/cmd/cover
-  - go get github.com/axw/gocov/gocov
```
adamdubiel: i removed those "magic" installs as they don't seem to be used anymore

Makefile
```diff
@@ -1,40 +1,62 @@
...
+
+TESTARGS ?= -race
+
+CURRENTDIR = $(shell pwd)
+SOURCEDIR = $(CURRENTDIR)
 SOURCES := $(shell find $(SOURCEDIR) -name '*.go')
 APP_SOURCES := $(shell find $(SOURCEDIR) -name '*.go' -not -path '$(SOURCEDIR)/vendor/*')
-VERSION=$(shell cat .goxc.json | python -c "import json,sys;obj=json.load(sys.stdin);print obj['PackageVersion'];")
+
+PATH := $(CURRENTDIR)/bin:$(PATH)
```
adamdubiel: setting this path once for all tests, thanks to running builds with consul on travis we got +10% to coverage for free ;)

Makefile
```diff
@@ -1,40 +1,62 @@
...
+
+COVERAGE_TARGETS = $(PACKAGES)
 
 all: build
 
 deps:
 	@./install_consul.sh
 
-build: deps test
+build-deps: deps format test check
```
adamdubiel: build runs `goimports` before starting build

janisz:
There is a problem with tarvis build. I restarted it but it failed again.

adamdubiel:
...and now it passed. Either we get back to not running `consul` tests with Consul on CI right now, or we wait to see how reliable this tests are in longer run, what do you think?

Makefile
```diff
@@ -1,40 +1,62 @@
...
+	go test -coverprofile=coverage/$(shell basename $@).coverprofile $(TESTARGS) $@
 
-gometalint-exists:
+check-deps: deps
 	@which gometalinter > /dev/null || \
-	(echo >&2 "gometalinter must be installed on the system. See https://github.com/alecthomas/gometalinter")
+        (go get github.com/alecthomas/gometalinter && gometalinter --install)
 
-check: gometalint-exists $(SOURCES)
+check: check-deps $(SOURCES)
```
janisz: I think we should follow [standard targets](https://www.gnu.org/software/make/manual/make.html#index-check-_0028standard-target_0029) so check should do style checking and testing.

Makefile
```diff
@@ -1,40 +1,62 @@
...
-	PATH=$(CURRENT_DIR)/bin:$(PATH) go test $(PACKAGES) $(TESTARGS)
-	go vet $(PACKAGES)
+	go test $(PACKAGES) $(TESTARGS)
+
+coverage-deps: deps
+	@mkdir -p $(COVERAGEDIR)
+	@which gover > /dev/null || \
+        (go get github.com/modocache/gover)
+
+test-coverage: coverage-deps $(SOURCES) $(COVERAGE_TARGETS)
```
janisz: I'd prefer to keep one test target. I think `test-coverage` looks better then current `test`, the output is cleaner and gives information about coverage.

adamdubiel: okay, that's reasonable, it doesn't seem to be slower

Makefile
adamdubiel: ack

janisz:
I think we should merge it. I think build hang could be not related to changes you made.

adamdubiel:
I know its not related to changes, but we might not be happy with builds failing randomly because of consul.

adamdubiel:
@janisz fixed, now we have 2 targets: `test` (testing with coverage) and `check` (test + metalinter)

PR #160, opened 2017-01-26
Add support for consul SSL auth

mhamann:
Even though the docs indicate that this client supports passing in SSL client and CA certs, there wasn't actually any code using those values.

Now there is. :-)

### DIFF
--- a/consul/agents.go
+++ b/consul/agents.go
@@ -2,7 +2,9 @@ package consul
 
 import (
 	"crypto/tls"
+	"crypto/x509"
 	"errors"
+	"io/ioutil"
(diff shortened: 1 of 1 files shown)


janisz:
Hi @mhamann 

Thanks four PR. It looks great. Unfortunately we have some problems with our travis build:
> [curl: (35) Unknown SSL protocol error in connection to releases.hashicorp.com:443](https://travis-ci.org/allegro/marathon-consul#L188)

I want to fix CI first before merge.

janisz:
@mhamann Closing due to lack of activity.

PR #161, opened 2017-01-26
Remove unused Consul dependency.

janisz:
It turns out we don't need consul binary to run the tests.
Removed consul installation script and dependency in Makefile.


janisz:
It turns out consul is required on path. If it's missing tests will be skipped
```
server.go:167: consul not found on $PATH, skipping
```

PR #162, opened 2017-01-26
Fix consul curl issue on Travis

mhamann:
Testing a possible fix for the curl issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: go
+dist: trusty
+sudo: false
 
 go:
   - 1.7
--- a/install_consul.sh
+++ b/install_consul.sh
@@ -20,7 +20,7 @@ hash consul 2>/dev/null || {
         os="windows"
         arch="386"
     fi
-    version="0.7.2"
+    version="0.7.3"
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -1,4 +1,5 @@
 language: go
+dist: trusty
```
janisz: I think we need 
```yaml
sudo: false
dist: trusty
```

mhamann: The build made it much further, but hit another error.

janisz: It passed for me after adding sudo line  https://travis-ci.org/allegro/marathon-consul/builds/195645241

janisz:
@mhamann Would you like to update this PR?

mhamann:
Will do a bit later today. Thanks!

janisz:
For some reason build hangs. Maybe we should update consul to 7.3

mhamann:
The build hang seems to be intermittent. I did bump Consul to 0.7.3, but does not seem to have alleviated the problem.

Something is flaky...

janisz:
@mhamann Thanks!

PR #173, opened 2017-02-02
FIX *.locahost is resolvable with nss-myhostname

kamilchm:
nss-myhostname, libnss_myhostname.so.2 — Provide hostname resolution for the locally configured system hostname.
> The hostnames "localhost" and "localhost.localdomain" (as well as any hostname ending in ".localhost" or ".localhost.localdomain") are resolved to the IP addresses 127.0.0.1 and ::1.
https://www.freedesktop.org/software/systemd/man/nss-myhostname.html


janisz:
@kamilchm Thanks!

PR #182, opened 2017-02-17
Fixes #181 | Count sync (de)registration

### LINKED ISSUE #181: Count sync (de)registration
We should collect how many (de)registration are performed by sync. If it's more than 0 then probably something went wrong.

tomez:


sync/sync.go
```diff
@@ -157,7 +165,9 @@ func (s *Sync) registerAppTasksNotFoundInConsul(marathonApps []*apps.App, servic
 					err := s.serviceRegistry.Register(&task, app)
 					if err != nil {
 						log.WithError(err).WithField("Id", task.ID).Error("Can't register task")
+						continue
```
kamilchm: I would prefer incrementing counter in else condition

tomez: Fixed

PR #190, opened 2017-02-27
Recheck if service should be deregistered 

janisz:
When service is marked to be deregisterd during sync ask marathon for
tasks to recheck if this task was not started after sync procedure started.
Prevents deregistartion of tasks started after sync asked for all
marathon services. It also minimize call to marathon for a price
of not deregistering tasks that was killed during sync. This will
be handled by events or next sync.

### DIFF
--- a/sync/sync.go
+++ b/sync/sync.go
@@ -124,24 +124,37 @@ func (s *Sync) resolveHostname() error {
 func (s *Sync) deregisterConsulServicesNotFoundInMarathon(marathonApps []*apps.App, services []*service.Service) (deregisterCount int, errorCount int) {
 	runningTasks := marathonTaskIdsSet(marathonApps)
 	for _, service := range services {
-		taskIDInTag, err := service.TaskId()
-		taskIDNotFoundInTag := err != nil
-		if taskIDNotFoundInTag {
+		logFields := log.Fields{
+			"Id":      service.ID,
+			"Address": service.RegisteringAgentAddress,
+			"Sync":    true,
+		}
+		if taskIDInTag, err := service.TaskId(); err != nil {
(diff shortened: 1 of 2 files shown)


sync/sync_test.go
```diff
@@ -395,6 +395,29 @@ func TestSync_WithDeregisteringProblems(t *testing.T) {
 	assert.Len(t, services, 1)
 }
 
+func TestSync_WithDeregisteringFallback(t *testing.T) {
```
adamdubiel: we should check what happens if marathon returns error

janisz: Fixed

sync/sync.go
```diff
@@ -211,3 +224,12 @@ func marathonTaskIdsSet(marathonApps []*apps.App) map[apps.TaskID]struct{} {
 	}
 	return tasksSet
 }
+
+func findTask(id apps.TaskID, tasks []*apps.Task) bool {
```
wendigo: maybe implement this in apps.

```
fun (tasks []*Task) FindById(id TaskID)
```

janisz: :+1: we could then fix it here [web/event_handler.go:176]https://github.com/allegro/marathon-consul/blob/c96bbf9b4fd5c33bc73b4b097844a43edc256cb4/web/event_handler.go#L176

janisz: This change will require to change `Marathoner.Tasks` answers from `[]*Task` to `[]Task`. I think it's good change since Task is small object (112 bytes) mostly pointers (4 strings and 2 arrays) so there should not be any problems. This change could be quite big and in multiple packages so I'd like to make it in separate PR #192

Issue #196, opened 2017-03-03
go get github.com/allegro/marathon-consul is failing


guilhem:
```
# cd /home/guilhem/go/src/github.com/allegro/marathon-consul; git submodule update --init --recursive
fatal: No url found for submodule path 'vendor/github.com/armon/go-metrics' in .gitmodules
package github.com/allegro/marathon-consul: exit status 128
```

janisz:
Related to https://github.com/Masterminds/glide/issues/683

janisz:
Fixed
```
➜  ~ export GOPATH=/tmp/gopath                                   
➜  ~ mkdir -p $GOPATH
➜  ~ go get github.com/allegro/marathon-consul
```

PR #197, opened 2017-03-03
Remove all submodules

guilhem:
this is fixing #196


janisz:
@guilhem Thanks for reporting. We are using `glide` to update vendor dir. Can you provide command you use to delete this sub modules? Any idea why is this happening?

guilhem:
@janisz this is because `glide install` used to create submodules (you have to use `--strip-vcs` to remove submodules)

Current `glide install` has:
```
[WARN]  The --strip-vcs flag is deprecated. This now works by default.
```

to remove submodules I have to `git rm --cached -r vendor` and `glide install` again

So, this problem should not happen again :)

janisz:
@guilhem Thanks

PR #201, opened 2017-03-07
Log error wehn events queue if full

janisz:


sse/sse_handler.go
```diff
@@ -90,22 +89,23 @@ func (h *SSEHandler) handle() {
 	delay := time.Now().Unix() - e.Timestamp.Unix()
 	metrics.UpdateGauge("events.read.delay.current", delay)
 	if e.Type != events.StatusUpdateEventType && e.Type != events.HealthStatusChangedEventType {
-		drop(fmt.Errorf("%s is not supported", e.Type))
+		log.Debugf("%s is not supported", e.Type)
```
janisz: We can log here error since this should never happen if we use marathon event type filtering.

PR #204, opened 2017-03-08
Speed up consul tests

janisz:
Setting `raft_multiplier` to 1 and disable automatic checking
for security bulletins and new version releases.

Refs:

* https://www.consul.io/docs/agent/options.html#raft_multiplier
* https://github.com/pszymczyk/embedded-consul/issues/32
* https://github.com/hashicorp/consul/issues/2705#issuecomment-283674916

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ sudo: false
 
 go:
   - 1.8
+env:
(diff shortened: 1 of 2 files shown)


janisz:
It looks like it does not help.

PR #206, opened 2017-03-10
Fix marathon events delay, now fetched from event data

tomez:


apps/task.go
```diff
@@ -3,10 +3,13 @@ package apps
 import (
 	"encoding/json"
 	"strings"
+
+	"github.com/allegro/marathon-consul/time"
 )
 
 type Task struct {
 	ID                 TaskID              `json:"id"`
+	Timestamp          time.Timestamp      `json:"timestamp"`
```
janisz: Pleas add a comment to this. This field is not a part of a Marathon task object. It's only present in the event. We parse task status event to this struct but it may be misleading in the future.

janisz: Please add the thest where this Timestamp is set

web/web_handler.go
```diff
@@ -50,8 +50,6 @@ func (h *EventHandler) Handle(w http.ResponseWriter, r *http.Request) {
 		}
 
 		metrics.Mark("events.requests." + e.Type)
-		delay := time.Now().Unix() - e.Timestamp.Unix()
-		metrics.UpdateGauge("events.requests.delay.current", delay)
```
janisz: Why this metric is removed?

time/time.go
```diff
@@ -25,10 +25,26 @@ func (t Timestamp) MarshalJSON() ([]byte, error) {
...
+
 func (t *Timestamp) String() string {
 	return t.Format(time.RFC3339Nano)
 }
 
+func (t *Timestamp) Missing() bool {
+	return t.Unix() == (time.Time{}).Unix()
+}
+
+func StringToTimestamp(date string) Timestamp {
```
janisz: if this is used only in test then put it there.

time/time.go
```diff
@@ -25,10 +25,26 @@ func (t Timestamp) MarshalJSON() ([]byte, error) {
 	return json.Marshal(t.String())
 }
 
+func (t *Timestamp) Delay() int64 {
+	return time.Now().Unix() - t.Unix()
```
janisz: Prefer [`Time.Sub`](https://godoc.org/time#Time.Sub)

apps/task.go
tomez: @janisz maybe creating separate struct for parsing StatusUpdate event ?

web/web_handler.go
tomez: Similar is pushed from event_handler.go:106,156 based on event after event parsing now.
Keep it for clarity ?

apps/task.go
janisz: :+1:  Good idea.

time/time.go
tomez: Done

time/time.go
tomez: Done

apps/task.go
tomez: Done as sugested

apps/task_test.go
```diff
@@ -13,6 +14,7 @@ func TestParseTask(t *testing.T) {
 
 	testTask := &Task{
 		ID:                 "my-app_0-1396592784349",
+		Timestamp:          time.Timestamp{},
```
janisz: Still missing a test with defined timestamp.

events/task_health_change_test.go
```diff
@@ -3,12 +3,22 @@ package events
...
 	"encoding/json"
 	"testing"
+	gotime "time"
 
+	"github.com/allegro/marathon-consul/time"
 	"github.com/stretchr/testify/assert"
 )
 
+func StringToTimestamp(date string) time.Timestamp {
+	t, err := gotime.Parse("2006-01-02T15:04:05.000Z", date)
```
janisz: Use RFC3339Nano instead of `"2006-01-02T15:04:05.000Z"`

PR #210, opened 2017-03-14
support for a path in marathon location

strzelecki-maciek:
Hi, would you be open to adjusting the range of supported values of the marathon location parameter?

In order to join multiple marathons (e.g. per dev-team, per project) we are discovering their (semi-random) locations through linkerd (https://linkerd.io/documentation/).

This changes our --marathon-location value from the standard hostname:port into something like  --marathon-location=localhost:4141/ops-staging/marathon-devteam-1

Proposed fix checks, naively, for the existence of "path" and divides it into host  and url that is then prefixing the path param.

### DIFF
--- a/marathon/marathon.go
+++ b/marathon/marathon.go
@@ -212,10 +212,9 @@ func (m Marathon) url(path string) string {
 
 type params map[string][]string
 
-// urlWithQuery was altered to handle a case when marathon location has a path in addition
-// to the ususal host & port. This was necessary in order to accommodate linkerd
-// (https://linkerd.io/documentation/) service discovery of marathon instance
-// eg. --marathon-location=localhost:4141/ops-staging/marathon-devteam-1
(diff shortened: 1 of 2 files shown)


marathon/marathon.go
```diff
@@ -212,12 +212,27 @@ func (m Marathon) url(path string) string {
 
 type params map[string][]string
 
+// urlWithQuery was altered to handle a case when marathon location has a path in addition
```
janisz: Nice comment. I'd prefer to change it a little. Comment should describes what and why function is doing not what happen to the function. Try to fill following template:

```
urlWithQuery returns ...
If location is given with path e.g, .... then ...
```

strzelecki-maciek: Thanks! Will adjust the comment and provide tests soon(ish)

janisz:
@strzelecki-maciek Thanks!

PR #212, opened 2017-03-23
Fixes #209 | Sanitize http logs

### LINKED ISSUE #209: Santize logs
Currenlty logs can contain passwords to Marathon. 

```bash
Get http://***:***@localhost:8080/v2/leader: dial tcp 127.0.0.1:8080: getsockopt: connection refused
```

We should not log it and best keep it in header.

tomez:
Put HTTP authentication in Header instead of URL. This will prevent leaking credentials in logs when error is logged with authentication information.

### DIFF
--- a/marathon/marathon.go
+++ b/marathon/marathon.go
@@ -31,7 +31,8 @@ type Marathon struct {
 	Location string
 	Protocol string
 	MyLeader string
-	Auth     *url.Userinfo
+	username string
+	password string
 	client   *http.Client
 }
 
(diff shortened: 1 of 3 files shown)


marathon/marathon.go
```diff
@@ -181,6 +181,9 @@ func (m Marathon) get(url string) ([]byte, error) {
 	metrics.Time("marathon.get", func() { response, err = m.client.Do(request) })
 	if err != nil {
 		metrics.Mark("marathon.get.error")
+		// error passed from http client could contain login and password
+		// this part is  muting it
+		err = fmt.Errorf("HTTP error while GET. Host: %s, URI: %s", request.Host, request.URL.RequestURI())
```
janisz: I think we should keep response code here

tomez: If we encounter error here then response is nil so there is no status.

PR #214, opened 2017-03-30
Decrease request retry number to speed up tests

misiek08:
On OS X executing tests with request retry number set to 100 was taking too much time. Now it finishes in 25-60s which is acceptable.


janisz:
We should add CI for OS X.

misiek08:
Should I add this in this PR or make another?

janisz:
@misiek08 Thanks!

PR #215, opened 2017-03-30
Enable tag override to allow extrnal tools adding tags

misiek08:
In some cases we want to add tags to services already registered in consul. This option allow external tools to add tags to existing services and order Consul to keep tags and propagate to cluster.

### DIFF
--- a/consul/consul_test.go
+++ b/consul/consul_test.go
@@ -133,6 +133,32 @@ func TestGetService_FailingAgent_GivingUp(t *testing.T) {
 	assert.Nil(t, services)
 }
 
+func TestServiceHasEnabledTagOverriding(t *testing.T) {
+	t.Parallel()
+
+	server := CreateTestServer(t)
+	defer server.Stop()
+
+	consul := ClientAtServer(server)
+
+	app := utils.ConsulApp("serviceA", 1)
+	app.Tasks[0].Host = server.Config.Bind
+	app.Labels["test"] = "tag"
+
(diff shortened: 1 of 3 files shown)


service/service.go
```diff
@@ -19,6 +19,7 @@ type Service struct {
 	Name                    string
 	Tags                    []string
 	RegisteringAgentAddress string
+	EnableTagOverride 		bool
```
janisz: Is this go formated?

consul/consul_test.go
```diff
@@ -133,6 +133,32 @@ func TestGetService_FailingAgent_GivingUp(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	consul_services, err := consul.GetServices("serviceA")
+	services := make(map[string]*service.Service)
+	for _, s := range consul_services {
+		services[s.Name] = s
+	}
+
+	assert.NoError(t, err)
+	assert.Equal(t, true, services["serviceA"].EnableTagOverride)
```
janisz: `assert.True`

janisz:
I merged this PR but I'm not sure if it's valid and wanted. Why do we want to disable anty entropy mechanism for all services?

> [The enableTagOverride can optionally be specified to disable the anti-entropy feature for this service.](https://www.consul.io/docs/agent/services.html#service-definition)

@misiek08 @adamdubiel Can you explain why we need this and what are the consequences of using this feature?

PR #232, opened 2017-05-19
Substitute $HOST and $PORT* in healthcheck command

guilhem:


consul/consul_test.go
```diff
@@ -1001,3 +1001,47 @@ func TestMarathonTaskToConsulServiceMapping_NotResolvableTaskHost(t *testing.T)
...
+					Host:  "127.0.0.6",
+					Ports: []int{8090, 8443},
+				},
+			},
+			want: "true",
+		},
+		{
+			name: "with ports and host",
+			args: args{
+				s: "nc $HOST $PORT0 && nc $HOST $PORT1",
```
janisz: I know this is rare scenarion but what if command has escaped variable name? e.g.: `echo \$HOST` or `echo '$HOST'`

guilhem: Yes that can be a problem... but solving this can be very difficult. I found this package https://godoc.org/github.com/mvdan/sh/syntax to parse and "understand" shell but API is not simple.

guilhem:
I don't know in case of bridged network... I just sure that it doesn't work for them either in current situation ;)

Marathon is running healthcheck shell commands with same execution context than a task... it's really hard to reproduce it.

consul/consul_test.go
janisz: @tomez @dankraw What do you think about it. I'm fine with this simple approach if we document it.

guilhem: @janisz documentation done ;)

janisz:
You are right. This feature is not working right now so adding it should not damage any Marathon-Consul installation.

consul/consul_test.go
tomez: Looks good to me.

janisz:
@guilhem Thanks!

PR #233, opened 2017-05-19
Disable Leader detection with a '*'

guilhem:
This is useful if you always have only 1 instance running (think of `marathon-consul` as a marathon app ;) ) and marathon address is provided by consul DNS.

Do we missed something?

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,7 +215,7 @@ marathon-protocol           | `http`          | Marathon protocol (http or https
 marathon-ssl-verify         | `true`          | Verify certificates when connecting via SSL
 marathon-timeout            | `30s`           | Time limit for requests made by the Marathon HTTP client. A Timeout of zero means no timeout
 marathon-username           |                 | Marathon username for basic auth
-marathon-leader             |                 | Marathon cluster-wide node name (defaults to <hostname>:8080), the some leader specific calls will be made only if the specified node is the current Marathon-leader")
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -215,7 +215,7 @@ marathon-protocol           | `http`          | Marathon protocol (http or https
 marathon-ssl-verify         | `true`          | Verify certificates when connecting via SSL
 marathon-timeout            | `30s`           | Time limit for requests made by the Marathon HTTP client. A Timeout of zero means no timeout
 marathon-username           |                 | Marathon username for basic auth
-marathon-leader             |                 | Marathon cluster-wide node name (defaults to <hostname>:8080), the some leader specific calls will be made only if the specified node is the current Marathon-leader")
+marathon-leader             | `<hostname>:8080` | Marathon cluster-wide node name, the some leader specific calls will be made only if the specified node is the current Marathon-leader. Set to `*` to always act like a Leader.
```
janisz: This part of readme is copied from [`config/config.go`](https://github.com/allegro/marathon-consul/blob/master/config/config.go) could you update description there too? We should make it autogenerated.

guilhem: oops ok :)

guilhem: I understand why there is no value for "default"

guilhem: Done

janisz:
@guilhem Thanks!

Issue #170, opened 2017-02-02
Use app definition for task in same version.


janisz:
With https://github.com/allegro/marathon-consul/pull/145 and https://github.com/allegro/marathon-consul/issues/164 we put a plaster on Marathon app task differences. We should obtain app definition for task in version defined in task. We can use `/v2/apps/{app_id}/versions/{version}` endpoint to get this information. What's more we can cache app definitions and to reduce calls to marathon. Only drawback of this solution is versions could not be available since it could be configured to keep last `n` entries. We need to think what we should do in that circumstances.

janisz:
App definition for given version is immutable so we can cache it. This will reduce Marathon queries but will increase memory usage (one app definition JSON takes approx 2.36K). We need to handle applications that was not changed but only scaled. This means we need to persist App Definitions. Cache  entry could be evicted if it's not used within sync interval. We could use https://github.com/golang/groupcache but then we need to think about data peristency if one or more instance goes down (e.g. version update).

PR #234, opened 2017-05-19
Do not manage events on non-consul apps

guilhem:
This fix a lot of issues when dealing with apps not managed by marathon-consul

### DIFF
--- a/events/event_handler.go
+++ b/events/event_handler.go
@@ -147,11 +147,12 @@ func (fh *eventHandler) handleHealthyTask(body []byte) error {
 }
 
 func (fh *eventHandler) handleStatusEvent(body []byte) error {
-	task, err := apps.ParseTask(body)
+	task, err := ParseTaskHealthChange(body)
 	if err != nil {
 		log.WithError(err).WithField("Body", body).Error("Could not parse event body")
 		return err
 	}
+
 	delay := task.Timestamp.Delay()
 	metrics.UpdateGauge("events.read.delay.current", int64(delay))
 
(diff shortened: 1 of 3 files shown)


events/event_handler.go
```diff
@@ -160,6 +161,18 @@ func (fh *eventHandler) handleStatusEvent(body []byte) error {
 		"TaskStatus": task.TaskStatus,
 	}).Info("Got StatusEvent")
 
+	app, err := fh.marathon.App(task.AppID)
```
janisz: This operation is quite expensive and could fail. Why you want to cut non consul app on this level?

guilhem: Right now, when a "Killing" or "Failed" event happen, it try to deregister application even if it's not consul.

We have a lot of `level=error msg="There was a problem deregistering task" Id={APP}.e41a91e6-3be7-11e7-bcb0-c4346bb6b3b0 error="Couldn't find any service matching task id {APP}.e41a91e6-3be7-11e7-bcb0-c4346bb6b3b0"`

guilhem:
@janisz I dig really into code and don't find why tests are failing...

events/event_handler.go
janisz: I need to test it, this could double the traffic to Marathon for a deployment. Becouse new tasks will need to be registered and old will check if should be deregistered. 

Instead of asking marathon I'll prefer to ask consul if it has specific task and then deregister it. Third solution is to check error and log if it's not error that means there is no such task. If we want to deregister task and it's already gone we are done and this is not an error.

janisz: I checked the code and we already did it. IMO best to change error and check if this is real error. Again if we want to deregister a service that is not registered this is not a problem. The problem was before, why we haven't registered it if we should?

guilhem: problem is that we have task managed by tool _mesos-consul_ when tag `consul` isn't present. and tasks with tag `consul` managed by _marathon-consul_ (as a transition step).

We really don't want to see _marathon-consul_ deregistering apps managed by _mesos-consul_.

janisz: Marathon-consul sets different service ID then mesos-consul so it won't delete apps registered by mesos-consul.

guilhem: you're right :)

Maybe a change on log is a good trade-off.
Best will be to create an internal "cached" list of applications that we managed (sync periodically).

janisz: I'm not sure if I understand. Caching applications/tasks that are manages could help with performance (less calls to Marathon and Consul) and can be first step to fix https://github.com/allegro/marathon-consul/issues/170

guilhem:
I'm closing this PR because as seen in comments it's not the right way to do it.

PR #236, opened 2017-05-26
default HTTP(s) Path to / if empty

tokenrain:
I have never programmed in Go. I am pretty sure this is the place for the change. Hopefully the syntax is correct for assigning to that variable.

### DIFF
--- a/consul/consul.go
+++ b/consul/consul.go
@@ -328,7 +328,11 @@ func (c *Consul) marathonToConsulChecks(task *apps.Task, healthChecks []apps.Hea
 
 		switch check.Protocol {
 		case "HTTP", "HTTPS", "MESOS_HTTP", "MESOS_HTTPS":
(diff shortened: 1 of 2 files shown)


consul/consul.go
```diff
@@ -326,6 +326,10 @@ func (c *Consul) marathonToConsulChecks(task *apps.Task, healthChecks []apps.Hea
 			Status:   "passing",
 		}
 
+                if check.Path == nil {
```
janisz: Replace `nil` with `""'`. It's not a pointer so it can't be `nil`.

tokenrain:
Update the nil to "" - thanks!

janisz:
I updated the PR with small test fixes. It looks like this behaviour was on purpose (https://github.com/allegro/marathon-consul/commit/65c47ca783776c510d318726065761eac2cf43cb) since we have test for not passing empty paths. 

Thanks again for help.

PR #247, opened 2017-07-18
Use new tags slice for every service.

janisz:


janisz:
Here is a blog post with motivation for this pull request (if tests are not enoguht) http://allegro.tech/2017/07/golang-slices-gotcha.html

letientai299:
I come here thanks to the blog post. Above example would be easier to grapsh if we add 2 new lines, as:

```go
package main

import (
	"fmt"
)

func a() {
	x := []int{0, 1}
	y := append(x, 2)
	z := append(x, 3)
	fmt.Println(y, z)

}

func b() {
	x := []int{0, 1}
	fmt.Println("cap(x) before:", cap(x))
	x = append(x, 2)
	fmt.Println("cap(x) after:", cap(x))
	y := append(x, 3)
	z := append(x, 4)
	fmt.Println(y, z)

}

func main() {
	a()
	b()
}
```
https://play.golang.org/p/RSEhh0COf8H

Issue #246, opened 2017-06-19
please add configurable number of retries/retry interval for marathon connection errors


hokiegeek2:
I marathon-consul gets a connection refused error from marathon, it currently restarts 4 times very quickly--1-2 seconds--which may not give marathon time to reset. A configurable number of marathon connection retries and retry time interval for marathon connection failures would (hopefully) account for situations where there are intermittent marathon connectivity issues

janisz:
It looks like there was intention to do it. SSE config has fields for configuring retries and backoff https://github.com/allegro/marathon-consul/blob/1.3.4/sse/config.go#L5-L6 this varaibles are set only when config is parsed from JSON. We should expose them in command line and describe in README.

PR #255, opened 2017-07-31
Allow configuration of SSE recover retries.

janisz:
Fixes: #246

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -87,6 +87,8 @@ func (config *Config) parseFlags() {
 	flag.Int64Var(&config.Web.MaxEventSize, "event-max-size", 4096, "Maximum size of event to process (bytes)")
 
 	// SSE
+	flag.IntVar(&config.SSE.Retries, "sse-retries", 0, "Number of times to recover SSE stream.")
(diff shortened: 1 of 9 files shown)


config/config.go
```diff
@@ -89,6 +89,8 @@ func (config *Config) parseFlags() {
 
 	// SSE
 	flag.BoolVar(&config.SSE.Enabled, "sse-enabled", false, "Enable sse event stream.")
+	flag.IntVar(&config.SSE.Retries, "sse-retries", 0, "Number of times to recover SSE stream.")
+	flag.IntVar(&config.SSE.RetryBackoff, "sse-retry-backoff", 0, "Configuration of initial time between retries to recover SSE stream.")
```
adamdubiel: should it be duration? or at least specify time unit?

janisz: Correct we should use duration here.

```
		seconds := time.Duration(i * s.retryBackoff)
		time.Sleep(seconds * time.Second)
```

https://github.com/allegro/marathon-consul/blob/e36ce2dc17344704e6afb217f5f103c83c585264/marathon/streamer.go#L68-L69

janisz: Fixed

hokiegeek2:
@janisz  @adamdubiel Thanks guys!!

PR #264, opened 2017-09-27
 support leader discovery logic so can support port other than 8080

tokenrain:
I have no go experience but I thought I would give this a shot. Basically we run Marathon in port 8800 and can not use the marathon-consul feature of having it sync only on the master because marathon-consul can only support finding masters if they run on port 8080 due to the hardcoding of that in the resolveHostname function. This may not be the exact code that you would use but I hope at least I have come close. Thanks!

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -94,6 +94,7 @@ func (config *Config) parseFlags() {
 	flag.BoolVar(&config.Sync.Enabled, "sync-enabled", true, "Enable Marathon-consul scheduled sync")
 	flag.DurationVar(&config.Sync.Interval.Duration, "sync-interval", 15*time.Minute, "Marathon-consul sync interval")
 	flag.BoolVar(&config.Sync.Force, "sync-force", false, "Force leadership-independent Marathon-consul sync (run always)")
+	flag.IntVar(&config.Sync.Port, "sync-port", 8080, "Marathon port (defaults to 8080)")
 
 	// Marathon
 	flag.StringVar(&config.Marathon.Location, "marathon-location", "localhost:8080", "Marathon URL")
--- a/marathon/marathon.go
+++ b/marathon/marathon.go
@@ -243,13 +243,13 @@ func (m Marathon) urlWithQuery(path string, params params) string {
 	return marathon.String()
 }
 
-func (m *Marathon) IsLeader() (bool, error) {
+func (m *Marathon) IsLeader(port uint32) (bool, error) {
 	if m.MyLeader == "*" {
 		log.Debug("Leader detection disable")
 		return true, nil
 	}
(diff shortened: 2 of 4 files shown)


adamdubiel:
Getting rid of hardcoded configuration options is always a good thing. However could you fix test code as well? Try `make build` on localhost and follow the output.

config/config.go
```diff
@@ -94,6 +94,7 @@ func (config *Config) parseFlags() {
 	flag.BoolVar(&config.Sync.Enabled, "sync-enabled", true, "Enable Marathon-consul scheduled sync")
 	flag.DurationVar(&config.Sync.Interval.Duration, "sync-interval", 15*time.Minute, "Marathon-consul sync interval")
 	flag.BoolVar(&config.Sync.Force, "sync-force", false, "Force leadership-independent Marathon-consul sync (run always)")
+	flag.IntVar(&config.Sync.Port, "sync-port", 8080, "Marathon port (defaults to 8080)")
```
janisz: There is no need to add the port configuration. We can parse port from `marathon-location` and use it in `IsLeader` function.

marathon/marathon.go
```diff
@@ -243,26 +243,26 @@ func (m Marathon) urlWithQuery(path string, params params) string {
...
 	return m.MyLeader == leader, err
 }
 
-func (m *Marathon) resolveHostname() error {
+func (m *Marathon) resolveHostname(port uint32) error {
 	hostname, err := hostname()
 	if err != nil {
 		return err
 	}
-	m.MyLeader = fmt.Sprintf("%s:8080", hostname)
```
janisz: Instead of `8080` there should be something like this

```go
u, err := url.Parse(m.Location)
if err != nil {
   return fmt.Errorf("Could not parse marathon location (%s): %v", m.Location, err);
}
m.MyLeader = fmt.Sprintf("%s:%d", hostname, u.Port())
```

janisz:
@tokenrain Good catch! Thanks! I have no idea how it could pass our review. I left some comments on the code. We definitely want to merge it.

janisz:
@tokenrain Would you like to work on this? I can fix issues for you if have no time.

tokenrain:
Sorry for the very late response. I was away and not focusing on technology stuff. I would be happy for you to take care of this as I am quite busy. Let me know if you can else I will try and get some time this week to take a stab. Many Thanks!

janisz:
No problem. I will take care of it. Thanks for update and once again thanks for catching this bug.

tokenrain:
No thank you for writing and maintaining such a valuable piece for software!!!!!

janisz:
I created separate PR for this #274

PR #268, opened 2017-10-09
Migrate glide to dep

### LINKED ISSUE #267: Migrate from Glide to dep
It looks like https://github.com/golang/dep is going to be a standard way of maintaining dependencies in Golang projects.

### TODO

- [ ] Genrate dep metada files (Gopkg.lock and Gopkg.toml)
- [ ] Delete glide's metadata files ([glide.lock](https://github.com/allegro/marathon-consul/blob/1.4.1/glide.lock) and [glide.yml](https://github.com/allegro/marathon-consul/blob/1.4.1/glide.yaml))

Refs:
* https://github.com/kubernetes-incubator/node-feature-discovery/pull/76
* https://github.com/coreos/docs/issues/775
* https://github.com/inadarei/justgo/issues/4

fortrieb:
Fix #267


Gopkg.toml
```diff
@@ -0,0 +1,55 @@
...
+#  name = "github.com/x/y"
+#  version = "2.4.0"
+
+
+[[constraint]]
+  name = "github.com/Sirupsen/logrus"
+  version = "0.11.5"
+
+[[constraint]]
+  name = "github.com/allegro/marathon-consul"
```
janisz: I think this should not be here

fortrieb: Project contains different packages and these are also dependencies to main program. That's because is also included in Gopkg.

fortrieb:
You are right it should not add new files to ``vendor/``, but dep tries to update dependency to the newest compatible version of a package. You can find detailed information about setup in [dep Readme](https://github.com/golang/dep#setup).

janisz:
New files are only generated by dependency to project itself. I've checked commit hashes and this change do not change used versions.

I updated linter so once you rebase build should pass.

Gopkg.toml
janisz: Take a [look at other projects](https://github.com/search?o=asc&q=filename%3AGopkg.toml&s=indexed&type=Code&utf8=%E2%9C%93) (e.g., https://github.com/golang/dep/blob/master/Gopkg.toml) you can mark your project as a dependency to itself.

Gopkg.lock
```diff
@@ -0,0 +1,116 @@
+# This file is autogenerated, do not edit; changes may be undone by the next 'dep ensure'.
+
+
+[[projects]]
+  name = "github.com/Sirupsen/logrus"
+  packages = ["."]
+  revision = "ba1b36c82c5e05c4f912a88eab0dcd91a171688f"
+  version = "v0.11.5"
+
+[[projects]]
```
janisz: Please delete this as well

Gopkg.toml
```diff
@@ -0,0 +1,51 @@
+
+# Gopkg.toml example
```
medzin: This comment block is unnecessary.

janisz:
@fortrieb Thanks!

janisz:
@fortrieb It looks like you deleted more lines than any other contributor :smile:  https://github.com/allegro/marathon-consul/graphs/contributors?from=2015-04-12&to=2017-10-11&type=d

Issue #271, opened 2017-10-17
Update Consul to 1.0.0


janisz:
We should do #270 first. Then release marathon-consul and finally perform same steps to update to 1.0. There might be some breaking changes so the code might need some fixes. Finally, we should check if Marathon-Consul is backwards compatible with Consul 0.9.3. And if everything is ok we may create a new release.

shayangz:
any plans on merging this PR and cutting a release?

Consul 1.1.0 [removed](https://github.com/hashicorp/consul/blob/master/CHANGELOG.md#110-may-11-2018) the deprecated `script` checks which is breaking marathon-consul checks.

pbetkier:
We're already on Consul 1.2.0 as of #288.

PR #275, opened 2017-10-19
Update Consul to 1.0.0 #271

silentsokolov:
Update Consul to 1.0.0


Makefile
```diff
@@ -31,6 +31,9 @@ deps:
         (go get github.com/laher/goxc)
 	@which goimports > /dev/null || \
         (go get golang.org/x/tools/cmd/goimports)
+	@which porter1 > /dev/null || \
```
janisz: What is `porter1` and why do we need it?

silentsokolov: Fix

janisz:
@silentsokolov Thanks! I added review comment. I'll be happy to merge it after you split it into two PR.

silentsokolov:
I do not think it's possible to split ... Porter released with consul 1.0 ((

janisz:
You are right. You install `porter` from repo but in code use vendored version. I'll create a release and test your chagnes. If ther will not be anny issues I'll merge it. But it may take over 1 week, defiently happen before end of month.

Alternatively you can remove deprecated RPC cofnig in tests. Then update Consul and finally use Porter. This will make commits fine grained and fixing one thing. What do you think?

silentsokolov:
I'll wait for you. I'm in no hurry to complete the hacktoberfest.

> You install porter from repo but in code use vendored version.

I view `porter` as an independent package. It is needed to run tests, but is not dependency. Or do not you think so?

janisz:
Using Gradle nomenclature it's a `testCompile` dependency. Without it, the test won't compile. But it's just a naming convention. Let's leave it as is and let me test it and merge it later. 
Thanks for your support.

pbetkier:
`porter` got replaced by `freeport` just 2 days later after this PR https://github.com/hashicorp/consul/pull/3605 :)

I will prepare another PR with update to Consul 1.2.0 and switching to `freeport` as a bonus.

Issue #276, opened 2017-10-30
Catalog get operation randomness problematic in heterogenous ACL environments


chemicL:
Currently, the read operation for list of services retries any cached agent in the cluster (obtained via local agent query for nodes).

The issue with that is in non-homogenous environments, where some agents are read-only and have no ACL configuration attached, and are handled by anonymous token.

When marathon-consul is equipped with a token, it can send it to a read-only agent, which will fail the request not knowing where to forward the ACL Token bound query.

Despite having a local agent defined (`consul-local-agent-host`), all read requests still go to a random node in the first place, which should be avoided, too.

Steps to reproduce:

1. Setup consul server cluster with ACL policies enabled (default policy allow),
2. Generate token for read operations for services,
3. Setup one consul client without ACL-DC configured,
4. Setup one consul client with ACL-DC pointing to the server cluster from step 1,
5. Configure marathon-consul on another node with a local consul client agent (it doesn't matter whether it's client has ACL-DC configured) and `consul-token` with the value from step 2.

Running marathon-consul in this setup should at some point yield logs:

    ... "error":"Unexpected response code: 403 (rpc error: rpc error: ACL not found)","level":"error","msg":"An error occurred getting services from Consul, retrying with another agent" ...

janisz:
Fixed by #277

PR #277, opened 2017-10-31
Local Agent used for service list reads (#276)

chemicL:
If defined, the local consul agent is used to read list of services.
Otherwise, a random agent is picked, but only from the cache of agents
already syncing registrations.
This change prevents situations, where contacting a random agent in the
cluster fails, because it has no ACL-DC configured.

### DIFF
--- a/consul/agents.go
+++ b/consul/agents.go
@@ -14,15 +14,17 @@ import (
 
 type Agents interface {
 	GetAgent(agentAddress string) (agent *Agent, err error)
+	GetLocalAgent() (agent *Agent, err error)
 	GetAnyAgent() (agent *Agent, err error)
 	RemoveAgent(agentAddress string)
 }
 
 type ConcurrentAgents struct {
-	agents map[string]*Agent
-	config *Config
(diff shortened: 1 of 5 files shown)


consul/agents.go
```diff
@@ -10,20 +10,21 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"github.com/allegro/marathon-consul/metrics"
 	"github.com/allegro/marathon-consul/utils"
-	consulapi "github.com/hashicorp/consul/api"
 )
 
 type Agents interface {
-	GetAgent(agentAddress string) (agent *consulapi.Client, err error)
+	GetAgent(agentAddress string) (agent *Agent, err error)
```
janisz: :+1:  For making `Agents` interface more consitent but please do it in a separate PR.

Kboxoxx:
> If defined, the local consul agent is used to read list of services. Otherwise, a random agent is picked, but only from the cache of agents already syncing registrations. This change prevents situations, where contacting a random agent in the cluster fails, because it has no ACL-DC configured.

PR #278, opened 2017-10-31
Agents interface unification

chemicL:
Change `GetAnyAgent` to return `Agent` instead of `Client` to make API consistent.


janisz:
@chemicL Thanks!

Issue #230, opened 2017-05-16
Support Marathon 1.5 networking API


janisz:
Marathon 1.5 will introduce new Networks API for apps. More information could be found:
* [Proposal](https://docs.google.com/document/d/1jTnuHLu6TVJ9FSLG7_L-ZMsUNlO-edSCwlHu3OVgANs/edit#)
* [Commit: 1fa0d9d1d](https://github.com/mesosphere/marathon/commit/1fa0d9d1d0e36e71f70efd83bd358fa8c10ecd9b)
* [Documentation PR](https://github.com/mesosphere/marathon/pull/5391)

zanes2016:
+1 this prevents us from using marathon-consul with Marathon 1.5+

janisz:
Oh,  we need to fix it shortly. The changes will not be big but we need to test it.

@zanes2016 Thanks for reminding me to work on this. Would you like to contribute a patch? I can give you any help you need.

zanes2016:
@janisz I won't have time in the next few weeks. Maybe someone else can take it up

zanes2016:
@janisz I'm looking into submitting a patch for this; however I'm running into issues building the binary from source:
```
$ make build
...
go test -coverprofile=coverage/marathon-consul.coverprofile -race _/Users/<user>/dev0/marathon-consul
can't load package: package _/Users/<user>/dev0/marathon-consul: cannot find package "_/Users/<user>/dev0/marathon-consul" in any of:
	/Users/<user>/dev0/marathon-consul/src/_/Users/<user>/dev0/marathon-consul (from $GOROOT)
	/Users/<user>/go/src/_/Users/<user>/dev0/marathon-consul (from $GOPATH)
make: *** [_/<path>/marathon-consul] Error 1
$
```
Any suggestions? This occurs on both OSX and Ubuntu 16.04.

janisz:
What Go version are you using? It looks like you do not have configured `GOHOME` nor `GOROOT` properly.

zanes2016:
You're correct, i don't have GOHOME or GOROOT set; but as long as GOPATH is set, it should work, no?
```
$ go version
go version go1.9.1 darwin/amd64
$ echo $GOPATH
/Users/dev/dev0/marathon-consul
$ echo $GOROOT

$ echo $GOHOME

```

zanes2016:
Aw... However, I do see it in the $GOBIN path:
```
$ echo $GOBIN/marathon-consul
[...]

PR #283, opened 2018-01-30
Support Marathon 1.5 Network API

zanes2016:
I've manually tested this end-to-end with Marathon, Consul, and this new binary. This patch addresses issue #230 

I'm able to manually compile this code using native `go run`. But I'm not able to run `make build`, etc. So maybe someone (@janisz) can test this?

### DIFF
--- a/apps/app.go
+++ b/apps/app.go
@@ -28,6 +28,10 @@ type PortDefinition struct {
 	Labels map[string]string `json:"labels"`
 }
 
+type Container struct {
+	PortMappings []PortDefinition `json:"portMappings"`
+}
+
 type appWrapper struct {
 	App App `json:"app"`
 }
@@ -37,6 +41,7 @@ type Apps struct {
 }
 
 type App struct {
+	Container       Container         `json:"container"`
 	Labels          map[string]string `json:"labels"`
 	HealthChecks    []HealthCheck     `json:"healthChecks"`
 	ID              AppID             `json:"id"`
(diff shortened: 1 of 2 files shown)


apps/app.go
```diff
@@ -162,14 +167,22 @@ type indexedPortDefinition struct {
 }
 
 func (app App) findConsulPortDefinitions() []indexedPortDefinition {
+	var appPortDefinitions []PortDefinition
+	if len(app.Container.PortMappings) > 0 {
```
janisz: Can you add a comment here with information about deprecation of `PortDefinitons` in favour of `PortMappings` and link to Marathon release notes?

janisz: We can extract this part to a function with proper godoc. e.g.

```go
func (app App) getPortDefinitions() []PortDefinition {
	var appPortDefinitions []PortDefinition
        if len(app.Container.PortMappings) > 0 {
		appPortDefinitions = app.Container.PortMappings
 	} else {
 		appPortDefinitions = app.PortDefinitions
 	}
        return appPortDefinitions
}
```

janisz: Can you add a test for this change?

zanes2016: Sounds good. Added the changes and unittest

apps/app.go
```diff
@@ -163,13 +168,28 @@ type indexedPortDefinition struct {
...
 		}
 	}
+
 	return definitions
 }
+
+// Deprecated: Allows for backward compatibility with Marathons' network API
+// PortDefinitions are deprecated in favor of Marathons' new PortMappings
+// see https://github.com/mesosphere/marathon/pull/5391
+func extractPortDefinitions(app *App) []PortDefinition {
```
janisz: Why not a struct receiver? 

`func (app App) extractPortDefinitions()`

zanes2016: Good suggestion. I'll update.

janisz:
@zanes2016 Thanks a lot!

zanes2016:
@janisz No problem! Thanks for the quick feedback

PR #297, opened 2019-11-08
don't search in all DCs if configured

ojagodzinski:
If some datacenter is marked in Consul cluster as "failed" Marathon-Consul still asks for services and fails after getting 500 in response. This PR adds option to configure single datacenter to ask for.

### DIFF
--- a/consul/consul.go
+++ b/consul/consul.go
@@ -64,7 +64,7 @@ func (c *Consul) getServicesUsingProviderWithRetriesOnAgentFailure(provide Servi
 }
 
 func (c *Consul) getServicesUsingAgent(name string, agent *consulAPI.Client) ([]*service.Service, error) {
-	dcAwareQueries, err := dcAwareQueriesForAllDCs(agent, c.config.Dc)
+	dcAwareQueries, err := dcAwareQueries(agent, c.config.Dc)
 	if err != nil {
 		return nil, err
 	}
@@ -80,11 +80,11 @@ func (c *Consul) getServicesUsingAgent(name string, agent *consulAPI.Client) ([]
 	return allServices, nil
 }
 
-func dcAwareQueriesForAllDCs(agent *consulAPI.Client, dc string) ([]*consulAPI.QueryOptions, error) {
-	if dc != "" {
+func dcAwareQueries(agent *consulAPI.Client, singleDc string) ([]*consulAPI.QueryOptions, error) {
+	if singleDc != "" {
 		var queries []*consulAPI.QueryOptions
 		queries = append(queries, &consulAPI.QueryOptions{
-			Datacenter: dc,
+			Datacenter: singleDc,
 		})
 		return queries, nil
 	}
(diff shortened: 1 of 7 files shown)


pbetkier:
I feel that [this](https://github.com/allegro/marathon-consul/pull/297/files#diff-e0cd2c42b7de4225cc3b6c0268572906L113) and [this](https://github.com/allegro/marathon-consul/pull/297/files#diff-e0cd2c42b7de4225cc3b6c0268572906L113) are the problems. The fact that we fail on first error instead of continuing with services that we were able to load. Querying only single DC feels more like a nice addition.

consul/consul.go
```diff
@@ -80,12 +80,18 @@ func (c *Consul) getServicesUsingAgent(name string, agent *consulAPI.Client) ([]
 	return allServices, nil
 }
 
-func dcAwareQueriesForAllDCs(agent *consulAPI.Client) ([]*consulAPI.QueryOptions, error) {
+func dcAwareQueriesForAllDCs(agent *consulAPI.Client, dc string) ([]*consulAPI.QueryOptions, error) {
+	if dc != "" {
```
AlfredBroda: This `dc` parameter shares a name with a variable on line 96 - this could lead to confusion, make it more unique perhaps?

consul/consul.go
```diff
@@ -110,7 +116,8 @@ func (c *Consul) getAllServices(agent *consulAPI.Client) ([]*service.Service, er
 	for _, dcAwareQuery := range dcAwareQueries {
 		consulServices, _, err := agent.Catalog().Services(dcAwareQuery)
 		if err != nil {
-			return nil, err
+			log.WithError(err).Error("An error occurred getting services from Consul, will continue with another DC")
```
AlfredBroda: Does it actually check another DC if there is only one DC query (when a DC is specified in `dcAwareQueriesForAllDCs`)? Same goes for line 251. BUT just change the text if checking this involves too much overhead.

ojagodzinski: there is always for loop over all DCs that Consul knows about (variable dcAwareQueries) https://github.com/allegro/marathon-consul/blob/consuldatacenters/consul/consul.go#L91 since Marathon in most of the deployements is not stretched across multiple DCs this logic was a bit wrong from the start ;)

consul/consul.go
ojagodzinski: fixd

consul/consul.go
```diff
@@ -110,7 +116,8 @@ func (c *Consul) getAllServices(agent *consulAPI.Client) ([]*service.Service, er
...
+			continue
 		}
 		for consulService, tags := range consulServices {
 			if contains(tags, c.config.Tag) {
```
pbetkier: still, if we cannot load one service we should continue imo, logging an error (line 119/126)

consul/consul.go
pbetkier: Do we add a metric for that?

consul/consul.go
```diff
@@ -80,12 +80,18 @@ func (c *Consul) getServicesUsingAgent(name string, agent *consulAPI.Client) ([]
 	return allServices, nil
 }
 
-func dcAwareQueriesForAllDCs(agent *consulAPI.Client) ([]*consulAPI.QueryOptions, error) {
+func dcAwareQueriesForAllDCs(agent *consulAPI.Client, singleDc string) ([]*consulAPI.QueryOptions, error) {
```
pbetkier: Would be more clear if it was:

```func dcAwareQueries(agent *consulAPI.Client, singleDc string) ... {
    if singleDc != "" {
        return dcAwareQueriesForSingleDC(...)
    }
    return dcAwareQueriesForAllDCs(...)
}
```

consul/consul.go
ojagodzinski: but there is continue statement

Issue #296, opened 2019-10-10
Task ID wrong in health status event


bartjolenovation:
Hi,

we are looking to upgrade marathon, and also marathon consul.
We are testing with Version: 1.4.2 (from apt) and marathon 1.9.

We see that when an application is stopped (by scaling to 0 instances) and then starting by setting 1 instance, the application is not registered in consul (the unregister works fine though). We see the following in the marathon-consul log:

`grep b98f6bf3-eb61-11e9-af4c-02423f927891 /var/log/marathon-consul.log
time="2019-10-10T13:27:38Z" level=info msg="Got StatusEvent" Id=demo_hello-world.instance-b98f6bf3-eb61-11e9-af4c-02423f927891._app.1 TaskStatus="TASK_STARTING" 
time="2019-10-10T13:27:38Z" level=debug msg="Not handled task status" Id=demo_hello-world.instance-b98f6bf3-eb61-11e9-af4c-02423f927891._app.1 taskStatus="TASK_STARTING" 
time="2019-10-10T13:27:38Z" level=info msg="Got StatusEvent" Id=demo_hello-world.instance-b98f6bf3-eb61-11e9-af4c-02423f927891._app.1 TaskStatus="TASK_RUNNING" 
time="2019-10-10T13:27:38Z" level=debug msg="Not handled task status" Id=demo_hello-world.instance-b98f6bf3-eb61-11e9-af4c-02423f927891._app.1 taskStatus="TASK_RUNNING" 
time="2019-10-10T13:27:41Z" level=info msg="Got HealthStatusEvent" Id=demo_hello-world.b98f6bf3-eb61-11e9-af4c-02423f927891 
time="2019-10-10T13:27:41Z" level=error msg="Task not found" Id=demo_hello-world.b98f6bf3-eb61-11e9-af4c-02423f927891 
time="2019-10-10T13:27:41Z" level=info msg="Got HealthStatusEvent" Id=demo_hello-world.b98f6bf3-eb61-11e9-af4c-02423f927891 
time="2019-10-10T13:27:41Z" level=error msg="Task not found" Id=demo_hello-world.b98f6bf3-eb61-11e9-af4c-02423f927891 `

So the "Got HealthStatusEvent" registers the following instanceID: demo_hello-world.b98f6bf3-eb61-11e9-af4c-02423f927891 where we see that the instance ID is demo_hello-world.instance-b98f6bf3-eb61-11e9-af4c-02423f927891 

[...]

PR #298, opened 2019-11-14
Change FindTaskById function

JurrianFahner:
This change will fix issue: https://github.com/allegro/marathon-consul/issues/296

It makes the FindTaskById function to match also on the first part of an id, so marathon-consul will be compatible with version 1.9 of marathon.

To avoid matching on too little information, also the requirement of at least a length of an id with 36 chars.

### DIFF
--- a/apps/task.go
+++ b/apps/task.go
@@ -44,7 +44,7 @@ type TasksResponse struct {
 
 func FindTaskByID(id TaskID, tasks []Task) (Task, bool) {
 	for _, task := range tasks {
-		if strings.HasPrefix(task.ID.String(), id.String()) && len(id.String) > 36 {
+		if strings.HasPrefix(task.ID.String(), id.String()) && (len(id.String()) > 36) {
 			return task, true
 		}
 	}
--- a/apps/task.go
+++ b/apps/task.go
@@ -44,7 +44,7 @@ type TasksResponse struct {
 
 func FindTaskByID(id TaskID, tasks []Task) (Task, bool) {
 	for _, task := range tasks {
-		if task.ID == id {
+		if strings.HasPrefix(task.ID.String(), id.String()) && len(id.String) > 36 {
 			return task, true
 		}
 	}
--- a/apps/task.go
+++ b/apps/task.go
@@ -44,11 +44,18 @@ type TasksResponse struct {
 }
 
 func FindTaskByID(id TaskID, tasks []Task) (Task, bool) {
+
+	if id.String() == "" {
+		// fail fast
+		log.Warning("Search for an empty string is not allowed")
+		return Task{}, false
+	}
+
 	var foundTask Task
 	amountOfTimesItHasBeenFound := 0
 
 	for _, task := range tasks {
-		if strings.HasPrefix(task.ID.String(), id.String()) && len(id.String()) > 1 {
+		if strings.HasPrefix(task.ID.String(), id.String()) {
 			foundTask = task
 			amountOfTimesItHasBeenFound++
 		}
--- a/apps/task.go
+++ b/apps/task.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/allegro/marathon-consul/time"
+	log "github.com/sirupsen/logrus"
 )
 
 type Task struct {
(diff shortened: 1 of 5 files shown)


apps/task.go
```diff
@@ -44,7 +44,7 @@ type TasksResponse struct {
 
 func FindTaskByID(id TaskID, tasks []Task) (Task, bool) {
 	for _, task := range tasks {
-		if task.ID == id {
+		if strings.HasPrefix(task.ID.String(), id.String()) && len(id.String) > 36 {
```
ojagodzinski: id.String -> id.String()

apps/task_test.go
```diff
@@ -41,15 +41,15 @@ func TestParseTasks(t *testing.T) {
 
 	expectedTasks := []Task{
 		{
-			ID:                 "test.47de43bd-1a81-11e5-bdb6-e6cb6734eaf8",
+			ID:                 "test.47de43bd-1a81-11e5-bdb6-e6cb6734eaf8._app.1",
```
ojagodzinski: Will it be backwards compatible with Marathon <1.9?

apps/task.go
JurrianFahner: I already saw it from the failing build, it was a last minute fix (and not doing the test locally).... 😬

apps/task_test.go
JurrianFahner: I've added an extra test, where the ids exactly match from event / task. See TestFindTaskByIdExactMatchBeforeMarathonVersionOneDotNine

JurrianFahner:
@ojagodzinski Is it possible to restart the build on the new commit?

ojagodzinski:
> @ojagodzinski The build is failing, how can it fixed? See below output of Travis. The package build succeeds now.
> 
> ```
> $ goveralls -coverprofile=coverage/gover.coverprofile -service travis-ci
> Error parsing coverage: open coverage/gover.coverprofile: no such file or directory
> The command "goveralls -coverprofile=coverage/gover.coverprofile -service travis-ci" exited with 1.
> Done. Your build exited with 1.
> ```

I enabled tests and removed one that was failing. Running some of them is still better than not running anything ;)

JurrianFahner:
I don't understand why the test is failing, it works on my computer (after pulling the latest changes). There is no setup needed, I think.

@ojagodzinski Do you understand what is happening, based on your experience?

ojagodzinski:
> I don't understand why the test is failing, it works on my computer (after pulling the latest changes). There is no setup needed, I think.

That is probably some kind of race condition because my branch (without your changes, only travis fix commit) works fine: https://github.com/allegro/marathon-consul/commit/2936975a7ff1e68f118730f4348482607be17309 and your changes doesn't seem to be connected with this test failing

JurrianFahner:
> > I don't understand why the test is failing, it works on my computer (after pulling the latest changes). There is no setup needed, I think.
> 
> There is a problem with len(id.String() > 36 condition, test cases have much shorter ids

That explains a lot indeed! The reason I added the `len(id.String() > 36` is to add some safety to avoid problems when an ID is not resolved and becomes an empty string (it would match). Before my change it wouldn't be a problem, since the method matched on equality. Now we are matching on the start of the strings, because the last part can't be predicted. I've chosen for 36, because it is the length of an uuid  (which is an arbritary choice). 

In my opinion we have two solutions:

1. Refactor all other tests for ids, to make sure that each id has more than 36 characters. But that is only reasonable when we are sure there is no use-case for this software where the ids are smaller. This might be also the safest option.

1. Lower the length of the minimal id in a way that all tests would succeed, which is definitely faster to change and least invasive. But might add some insecurity, when the id of an event might match more than one task.  

@ojagodzinski What would you like me to do? I'll pick up the task on monday.

ojagodzinski:
> Refactor all other tests for ids, to make sure that each id has more than 36 characters. But that is only reasonable when we are sure there is no use-case for this software where the ids are smaller. 

It is the most dangerous one. We cannot force apps to use names with some arbitrary length.

> is to add some safety to avoid problems when an ID is not resolved and becomes an empty string

so it should fail if ID is not resolved.

> Lower the length of the minimal id in a way that all tests would succeed, which is definitely faster to change and least invasive. But might add some insecurity, when the id of an event might match more than one task.

divide this case into two separate ones and check id via regex not via HasPrefix in marathon >1.9

JurrianFahner:
I changed the code a bit, to make it possible to match on ids which are of length 1. 
Also the case when there is a search on an empty string is covered, by the code.

Downside of this approach that for each search the whole array of tasks needs to be searched, to add more safety. Before this change the search was stopped when a task was found.

JurrianFahner:
`make release check` works on my machine, I don't know why it fails on travis.

JurrianFahner:
@ojagodzinski How can we resolve the issues in the pull request? I can't reproduce the issues on the travis build.

